PR #5, opened 2020-01-16
Generic containers

arjenmarkus:
I have added a sample program to process the templates as defined via the examples (note: it is far from complete, but it should illustrate a possible solution). I think the code is ready for review. 
There ought to be a more textual description of the details in these examples. That will be a next step. So, please have a look.

### DIFF
--- a/scenarios/Generic_Containers/dictionary.f90
+++ b/scenarios/Generic_Containers/dictionary.f90
@@ -80,7 +80,7 @@ template dictionary
     type, extends(linked_list_def) :: dictionary_type
         type(key_value_type)   :: key_value
     contains
-        procedure :: add => add_key_value      !! Still t obe added!
+        procedure :: add => add_key_value      !! Still to be added!
         procedure :: get => get_value
         procedure :: has_key => has_item_key
     end type dictionary_type
--- a/scenarios/Generic_Containers/dictionary.f90
+++ b/scenarios/Generic_Containers/dictionary.f90
@@ -3,7 +3,7 @@
 !     Dictionaries are data structures that allow the user to retrieve pieces
 !     of data via a key (typically a character string)
 !
-!     Dictionaries ass intended here hold data of the same type, so that
+!     Dictionaries as intended here hold data of the same type, so that
 !     various specific types may be needed. If they were to hold a variety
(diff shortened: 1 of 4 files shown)


tclune:
@arjenmarkus  I do not want to read too much into your intent with the preprocessor, but also need to let you know a bit more about the subcommittee's current approach to generics that might have some bearing.     

First and foremost,  plenary completely rejected preprocessors as the implementation.   (This  of course is not to say that a preprocessor would not be useful in the interim to enable creation of "working examples.)

But another, more subtle, aspect of our current approach is to demonstrate how existing OO mechanisms in Fortran can be leveraged for specification of parameter types passed into a generic.   This work is under the guidance of Magne Haveraaen, a template researcher from Bergen, Norway who is donating his time in the hopes that Fortran "gets it right".        One important aspect of his approach is to show the vendors that implementation of generics is not nearly as hard as they might be thinking.   But it also enables important capabilities such as type-safety which C++ does not quite get right.    Note, generic containers are actually degenerate case in this approach as the methods of the contained objects are immaterial.

Under Magne's approach, we would not need a preprocessor until we start to discuss actual syntax.  Though my recent experience suggests that even valid Fortran code is likely to break extant compilers, where as generated (non-polymorphic) code is more likely to be actually compilable.

At the next meeting, I'll try to quietly inquire if a behind-the-scenes, flexible, source-to-source processor might be useful for testing/demonstrating ideas.

arjenmarkus:
@tclune, I am familiar with Magne's work (well, the paper he wrote with his direct colleague and Damian Rouson). The preprocessor I wrote is merely meant to test whether the syntax I devised works - nothing more serious than that ;). As a matter of fact, I wrote an article (to be published in Fortran Forum) that uses existing features to achieve generics. I think we are very close - I should re-read Magne's paper, as I forgot the details of what they propose.

tclune:
Glad we're all on the same page.  Cheers!

arjenmarkus:
Just committed another example - this time legal Fortran 2003 code: a generic linked list and two specialisations. It requires a bit more code than perhaps desirable, but it demonstrates that we are close. Not part of a pull request yet.

PR #6, opened 2020-01-17
Feature/block matrix multiply

tclune:


arjenmarkus:
Just a small remark: the #defiine statement in one of the source files seems superfluous. I see the construct is actually used elsewhere, though.

tclune:
Yeah - I was trying to be a "good" programmer and use PURE but had enough debugging issues that I wanted to be able to use PRINT.    Thought it might be clever to define a macro PURE that could be disabled via cmake.    But did that about the same time that I got the system working.   And then never cleaned up.

PR #14, opened 2020-10-21
Added new directory theoretical work.

tclune:
The ./scenarios direcory is appropriate for use cases, but ./theory is
intended for prototyping and exploring various issues in defining
generics capabilities.


certik (approved):
Awesome. This is precisely what we need to do to figure it out. Thank you @tclune. +1 to merge from me.

PR #15, opened 2020-10-22
Variants on GenericIntrinics

tclune:
Added V2 that uses parameterized derived types (little benefit), and
V3 that proffers what could be done if kind-type parameters could be
assumed in dummy arguments.


certik (approved):
Looks good!

PR #18, opened 2020-10-22
Add Haskell example to comparisons

everythingfunctional:
Create Haskell equivalent to the example already provided for C++, Rust and Go.


certik (approved):
This is excellent, thank you!

I went over it quickly, and it looks good. So let's merge it. I will actually play with this in Haskell later.

PR #26, opened 2020-11-19
Initial start on use case paper

tclune:
This paper is intended to provide traceability between formal
requirements and use cases. The current structure is just a suggestion
and should be debated in generics subgroup.

Only one use case currently has nontrivial content.  I reformatted a
previously submitted use case paper into latex.


certik (approved):
Looks good as a start.

PR #27, opened 2020-12-03
Add initial notes on a value system

gklimowicz:
These are my hastily drawn initial notes on value systems and values for generics.

### DIFF
--- a/Value System.md
+++ b/Value System.md
@@ -1,0 +1,85 @@
+# A Value System for evaluating alternatives
+
+This paper attempts to collect our thoughts on creating an overarching value system for evaluating generic programming alternatives.
+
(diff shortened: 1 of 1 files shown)


tclune:
I have a general feeling that "Utility" should be broken into 2 (maybe more) separate values.      E.g. it is not difficult to imagine an approach to generics that makes it easy to implement a generic `swap` that works on non-pointers.   And also easy to implement a generic `swap` that works on pointers.     But an approach that allows a single template to handle both cases is preferred.      

I do not claim any insight on how best to split this category.   Words that possibly address concerns such as mine above are  "Expressivity" or "Parsimony".

tclune (approved):

PR #29, opened 2021-02-04
Create an example of generic error handling

everythingfunctional:
This is an example of a generic, functional style error handling utility I'd like to be able to write.


tclune:
I have had similar thoughts, and this is far more likely than getting the vendors to provide us with a tuple return code.

I have some more half-baked thoughts on how this might be made even more convenient.   They probably are unworkable but I'll try to play a bit and get back to you.

tclune (approved):

PR #30, opened 2021-02-17
Convert Use Cases and Requirements from LaTeX to markdown

everythingfunctional:


tclune (approved):
We'll want to start digesting these into smaller pages soon.

arjenmarkus:
I have read the document on the use cases and found a smal flaw ;). The
example in the State of the Art section (point 4), is in fact implementable
in Fortran 2008, I mean:

sparse(i,j) = x

where sparse is not a regular array but some container. The feature you
need is called a pointer function in note N1891 by John Reid. Here is a
trivial though full example:

module myfunc
    implicit none

    real, dimension(10), target :: array
contains

function storage(key) result(loc)
    integer, intent(in) :: key
    real, pointer :: loc
    loc => array(key)
end function
end module myfunc

program test_myfunc
    use myfunc

    array = 0.0
    storage(3) = 1.0

    write(*,*) array
end program test_myfunc

It works with Intel oneAPI Fortran and gfortran 10.0.2 at the very least.

Op di 23 feb. 2021 om 15:49 schreef Tom Clune <notifications@github.com>:

tclune:
Arjen,

Thank you – this is an issue that I’ve been confused (and optimistic) about for some time.      On my 4th dive, I’ve finally found where that is added in the standard.

This will be very valuable in my FPP-based container “templates”.

Cheers,


  *   Tom

arjenmarkus:
You're welcome - I must admit I have not seen this used in any actual
source code yet, but it looks like a very useful feature for containers
like associative arrays. I modified my example a bit to use a string
instead of an integer argument and that worked fine as well :).

Op di 9 mrt. 2021 om 16:17 schreef Tom Clune ***@***.***>:

PR #32, opened 2021-06-07
Draft of use cases paper.

tclune:


J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+"Automatic Allocation on READ into ALLOCATABLE character or array". It
+also had almost four pages of comments, while no other feature had as
+much as a full page of comments. In spite of the obvious demand for
+this feature, its complexity made it inappropriate to include in
+Fortran 202X. Instead it was planned that preliminary work on this
+feature would be fitted into the schedule of work on Fortran 202X in
+the hope that this would allow the completion of the work during the
+development of Fortran 202Y.
+
+Any syntax suggestions in this paper are notional, and only for the
```
everythingfunctional, line 25:
```suggestion
Any syntax suggestions in this paper are notional, and only for
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+illustrative purposes.
+
+This paper uses the term 'template' to refer to a generic entity that
+can be 'instantiated' into concrete implementations using sets of
+input 'template parameters'. While there are other possible approaches
+to generic programming, (e.g., intelligent macros), the guidance from
+J3 in the formation of the generics subgroup was to pursue something
+that is at least roughly along the lines of the template approach in
+other languages.  We also note that previous work on Parameterized
+Modules easily fits within this intentionally vague definitions of
```
everythingfunctional, line 35:
```suggestion
Modules easily fits within the intentionally vague definition of
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+templates in this context.
+
+NOTE 1: This paper is intended to be a living document.  As we develop
+the F202Y generics features, these use cases will be used for
+traceability in the subsequent requirements paper.  If additional use
+cases are encountered, or existing ones require modifications, new
+versions of this use case paper will be submitted for vote.  And if
+any use previously approved use cases are subsequently shown to be
```
everythingfunctional, line 43:
```suggestion
any previously approved use cases are subsequently shown to be
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+      TYPE(T) :: tmp
+
+      tmp = a ! note the use of copy assignment operator '='
+      a = b
+      b = tmp
+
+   END SUBROUTINE SWAP
+
+A few variations on SWAP functionality are worth considering. These
+are primary intended as proxies for analogous issues that naturally
```
everythingfunctional, line 101:
```suggestion
A few variations of SWAP functionality are worth considering. These
are primarily intended as proxies for analogous issues that naturally
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+The SWAP instantiation above is designed to work with scalars, but the
+underlying algorithm is equally applicable to arrays of arbitrary (but
+consistent) shape.  Rank agnostic features planned for F202X are of
+significant value in this context, but still require a lengthy SELECT
+RANK construct.  It is desirable to provide simple means that enable
+developers to declare templates that are rank agnostic.  E.g., one
+approach could be to allow integer template parameters that allow the
+rank to be specified explicitly and then more easily combined with
+rank already-planned rank-agnostic features.
```
everythingfunctional, line 117:
```suggestion
already-planned rank-agnostic features.
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+
+Variant 2: Parameterized types
+
+The instantiation above works for types that lack is intended to work
+with types , but the underlying algorithm is equally applicable to
+parameterized types provide that the actual type parameters of both
+dummy arguments agree.  It is desirable to provide developers with a
+simple means to declare templates that work with arbitrary
+parameterized types subject to appropriate requirements.
```
everythingfunctional, line 127: This paragraph got a little jumbled. I expect it is saying that the new functionality must interoperate with existing parameterized derived types, but I can't quite make sense of it.

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+2.1.3 Sorting and Searching
+---------------------------
+
+Sorting
+
+While the need to sort a list of items is somewhat rarer in typical
+Fortran applications than in wider software communities, it
+nonetheless arises often enough to be a problem. It is typically
+applied to a rank one array, but sorting algoritms can usually be
+generalized to other categorise of containers with appropriate
```
everythingfunctional, line 216:
```suggestion
generalized to other categories of containers with appropriate
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+iterators, e.g. Lists and Vectors. A given sorting algorithm can
+generally be applied to any type that provides a comparison (< or >)
+operation, or equivalent binary logical predicate. With current
+Fortran capabilities, the algorithm must be re-implemented for each
+type.  We note however, that some simplification possible through the
```
everythingfunctional, line 221:
```suggestion
type.  We note however, that some simplification is possible through the
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+use of include files and newly the newly introduced TYPEOF.
```
everythingfunctional, line 222:
```suggestion
use of include files and the newly introduced TYPEOF.
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+  or as an explicit parameter to the generic with a defined interface;
+
+* and the ability to take one or more iterators as instantiation
+  parameters.
+
+It is often necessary to sort the same data according to different
+comparison criteria.  It is therefore desirable that the Fortran
+generics facility provide means to specify the comparison operation as
+a separate template parameter.  This functionality is less crucial
+here than it is for the containers use case above, as in many cases
```
everythingfunctional, line 245:
```suggestion
here than it is for the containers use case below, as in many cases
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+
+
+
+2.1.4 Complex numerical algorithms
+----------------------------------
+
+Many complex numerical algorithms are written for specific intrinsic
+types and kinds, but are actually applicable to a broader set of
+types.  The applicability of such algoritms is often narrower than the
```
everythingfunctional, line 310:
```suggestion
types.  The applicability of such algorithms is often narrower than
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+This algorithm can be generalized to types TA, TB, TC that satisfy
+certain properties:
+
+SUBROUTINE GENERIC_MATMUL(A,B,C)
+    INTEGER :: i, j, k
+    INTEGER :: m, n, p
+    TYPE(TA), INTENT(IN) :: A(m,n)
+    TYPE(TB), INTENT(IN) :: B(n,p)
+    TYPE(TC), INTENT(IN) :: C(m,p)
```
everythingfunctional, line 358:
```suggestion
    TYPE(TC), INTENT(OUT) :: C(m,p)
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+    
+
+    DO i = 1, m
+       DO j = 1, p
+          C = C%zero()
```
everythingfunctional, line 363:
```suggestion
          C(i,j) = C(i,j)%zero()
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+    * Type TC must support an operation that returns a generalized "0"
+
+    * Type TA must support (matrix) multiplication by type TB
+      returning an intermediate type V
+
+    * Type TC must support addition by type V returning intermediate
+      type U
+
+    * Type TC must support assignment to an intermediate type U
```
everythingfunctional, line 382:
```suggestion
    * Type TC must support assignment from an intermediate type U
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+Note 1: These intermediate types may be the same as any of TA, TB, and
+TC, but the algorithm here is intended to maximize generality.
+
+Note 2: The "scalar" multiplication of types TB and TB in the nested
```
everythingfunctional, line 387:
```suggestion
Note 2: The "scalar" multiplication of types TA and TB in the nested
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+the lines of code are infrastructure as-opposed to direct numerical
+calculation. Examples of such infrastructure include software layers
+that couple independently developed subsystems, frameworks for
+managing distributed parallelism, advanced I/O (e.g.,
+checkpoint/restart via NetCDF), etc. Generally these infrastructure
+layers evolve as an attempt to avoid code duplication as multiple
+parts of the system require similar functionality.
+
+Software "containers" are abstractions that enable aggregating groups
+of related entities for convenient and efficient access of hte
```
everythingfunctional, line 424:
```suggestion
of related entities for convenient and efficient access of the
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+contained items.  Many distinct varieties of software containers have
+been developed and used.  By providing canonical and reliable means to
+perform common operations, containers can greatly simplify
+implementation of higher level algorithms.  design and implementation
+of complex algorithms.
```
everythingfunctional, line 429:
```suggestion
the design and implementation
of higher level complex algorithms.
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+Queue, etc.  Unlike Array containers, these others ar generally more
+dynamic - growing in size as necessary when new elements are added to the
+container. Here we briefly describe some of the most common varieties
+of containers.
+
+We also note that one approach to supporting the use cases in this
+section would be to more directly add specific container varieties as
+1st order features of Fortran.  However, the wide range of container
+varieties compounded with the even wider range of potential internal
+representations and algorithms suggests instead that language should
```
everythingfunctional, line 462:
```suggestion
representations and algorithms suggests instead that the language should
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+given insertion, a new larger block is automatically allocated and
+existing data is transferred to the new storage.  (Block size is
+typically doubled in such cases leading to O(n) asymptotic complexity
+for inserting n items.)  Removal of items at other locations is
+supported, but the operation is inefficient.
+
+The need for vectors is often encountered when the number of items
+that match some criterion is not known in advance.  Lacking such
+functionality, Fortran codes often implement these scenarios with 2
+sweeps with nearly identical logic.  The first sweeep merely counts
```
everythingfunctional, line 485:
```suggestion
sweeps with nearly identical logic.  The first sweep merely counts
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+the number of matches and then allocates an array.  The second sweep
+then populates the array.  Use of containers makes such algorithms
+much simpler to understand in many respects.
+
+* As with arrays, vector require no operations on the contained type
```
everythingfunctional, line 490:
```suggestion
* As with arrays, vectors require no operations on the contained type
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+* Set containers require that supported types provide an ordering
+  operation.
+
+
+2.2.3 Associative array (aka Map or Dictionary)
+
+Map containers store elements that are formed by a combinations of
+(unique( keys and values, and thus generally involved 2 distinct
+types.  To some degree, these containers can be regarded of as a
```
everythingfunctional, line 511:
```suggestion
Map containers store elements that are formed by a combination of
unique keys and values, and thus generally involve 2 distinct
types.  To some degree, these containers can be regarded as a
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+   I = begin(C)
+   loop while I is not end(C)
+      e = get(I)
+      < do something with e >
+     next(I)
+   end loop
+
+It is desirable that the generics facility provides features which
+will allow container implementation to provide companion iterator
```
everythingfunctional, line 568:
```suggestion
will allow container implementations to provide companion iterator
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+their template requirements to potential users, but also verify that
+instantiations will produce correct code for all types that meet the
+requirements.  Here we envision the case that a developer might miss
+documenting some necessary operation in the middle of complex code.
+The developer tests the template with types that just happen to
+support that undocumented operation, but some end user is not so
+fortunate.
+
+At the same time, it is recognized that there is value to having
+enforcement of verification protections might be inconvenient during
```
everythingfunctional, line 730:
```suggestion
At the same time, while it is recognized that there is value to having
enforcement of verification, these protections might be inconvenient during
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+
+
+Straw votes:
+
+CONCEPTS-1: Should template developers be able to clearly and
+            concisely communicate the implied requirements on the
+            types used in the template?  (YES,NO,UNDECIDED,ABSTAIN)
+
+CONCEPTS-2: Should template developers be able to verify that their
+            templates do not use any detect unspecified implied
```
everythingfunctional, line 752:
```suggestion
            templates do not use any unspecified implied
```

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,760 @@
...
+            requirements on the types used in the templates?
+            (YES,NO,UNDECIDED,ABSTAIN)
+
+CONCEPTS-3: Should it be possible to encapsulate template requirements
+            for reuse in multiple contexts? (YES,NO,UNDECIDED,ABSTAIN)
```
everythingfunctional, line 757: This doesn't necessarily imply that they would be composable. We should have an additional straw vote on whether they should be composable.

everythingfunctional:
I did verify that your numbering is correct.

J3-Papers/Generics_Use_Cases.txt
tclune: This is the key capability that would enable such composition.    I debated saying something like 'named encapsulation', but thought this should be enough for us to wedge in the desired requirement.

If the encapsulations are named, what more do you think is necessary for composition?

tclune:
Thank you @everythingfunctional for proof reading!

everythingfunctional:
I don't think adding the sparse arrays example I came up with is necessary, and may in fact be more confusing than helpful. I think the block matrix example covers most of it, especially since you explicitly mentioned that the blocks themselves may be block matrices and invoke some sort of "recursive" feature of templates.

Also, the discussion of wanting to "compose requirements" covered my other desire. Unfortunately I still haven't come up with a use case to "require" it without being so complicated that it would be hard to understand, or would be easier to write in another way without templates.

J3-Papers/Generics_Use_Cases.txt
everythingfunctional: Just being able to name them isn't sufficient. We must explicitly say that it be possible to define new requirements in terms of existing ones. It would be entirely possible to design the semantics and syntax in such a way that enabling that in the future becomes much more difficult.

tclune: Naming is a specific mechanism for reuse, and Im trying as much as possible not to imply a particular mechanism.   E.g., maybe the requirements themselves are not named but there is some way to say "and use the requirements from template X parameter Y"  (probably not, but it's just the first thing that popped into my head).

And I don't really think we are talking about defining new requirements in terms of existing ones.  We are talking about _including_ existing requirements that have been defined elsewhere.    Or maybe I'm just splitting hairs.   I'll think some more and come back after I've dealt with easier issues.

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,757 @@
...
+assignment.
+
+Notionally, the instantiation of the template for a type T should be
+equivalent to:
+
+   SUBROUTINE SWAP(a, b)
+      TYPE(T), INTENT(INOUT) :: a
+      TYPE(T), INTENT(INOUT) :: b
+
+      TYPE(T) :: tmp
```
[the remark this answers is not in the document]
tclune: Good point.   And there is also `TYPEOF` which was in my original draft.  It was suggested that I delete that, so I did.

ASSOCIATE won't work in some of the variants, so I don't think I'll switch to that, but I otherwise would.

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,757 @@
...
+
+    DO i = 1, m
+       DO j = 1, p
+          C(i,j) = 0
+          DO k = 1, n
+             C(i,j) = C(i,j) + A(i,k)*B(k,j)
+	  END DO
+       END DO
+    END DO
+
```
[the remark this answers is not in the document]
tclune: I think that the use of SUM also constrains the types that can be supported in a subtle way.  I doubt it ever really matters.

But in the original version type TC must be addable to the type TA*TB.  In the SUM version   TA*TB must be addable to itself.  (ANd similar twists for the assignment.)

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,757 @@
...
+over to optimized libraries (e.g., LAPACK) for types that correspond
+to single and double precision reals and complex
+
+
+Straw votes:
+
+ALG-9: Should the Fortran generics facility provide sufficient
+       generality to define numerical algorithms such as the
+       generalized linear algebra described above?
+       (YES,NO,UNDECIDED,ABSTAIN)
```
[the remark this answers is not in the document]
tclune: Interesting.

J3-Papers/Generics_Use_Cases.txt
```diff
@@ -0,0 +1,757 @@
...
+              type parameters?  (YES,NO,UNDECIDED,ABSTAIN)
+
+CONTAINERS-5: Should Fortran's generics facilities provide
+              functionality that allows a well-written container to
+              handle pointers? (YES,NO,UNDECIDED,ABSTAIN)
+
+CONTAINERS-6: Should Fortran's generics facilities enable users of
+              relevant containers to indepndently specify a comparison
+              operation at instantiation?  (YES,NO,UNDECIDED,ABSTAIN)
+
```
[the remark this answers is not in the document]
tclune: I've added it back.

J3-Papers/Generics_Use_Cases.txt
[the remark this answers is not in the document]
tclune: Note that this also does not sidestep the need for a zero-initialization - it merely hides it a bit more.    For instance what if n=0.

J3-Papers/Generics_Use_Cases.txt
tclune: How about this wording?   "CONCEPTS-3: Should it be possible to encapsulate template requirements
            for reuse in a composable manner? (YES,NO,UNDECIDED,ABSTAIN)"

everythingfunctional: I think I'd be happy with that.

rouson (approved):

arjenmarkus:
I just finished reading yesterday's  text. As I am always a trifle confused about how these reviews work, I use the pedestrian method. 

Lines 122-123: 

The instantiation above works for types that lack is intended to work
      
      
        with types , but the underlying algorithm ...

This sentence is either mangled or I simply don't understand it.

Lines 170, 177: "instrinsic"  -> "intrinsic"

Line 193:  "capabilites" -> "capabilities"

Ad 2.1.2
It is discussed later on, but it seems quite important to be able to define the requirements on the types that can be passed to the template: what operations are required, what relations there are between different arguments or template parameters and types. (How is that done in C++?) 

Line 221: "some simplification possible" -> "some simplification is possible"
Line 222: "newly the newly" -> "the newly"

Ad 2.1.3
What would happen in a MERGE algorithm if the two collections are NOT sorted? Is that a runtime error? 

Line 379: "is not recursive" - do we not consider recursive templates? They seem to provide quite some power in C++ (besides undoubtedly quite some complications ;))

Line 384: period missing at the end of the sentence.

Line 405: "as-opposed" -> "as opposed"

Line 418: "higher level algorithms." - period should be a comma

Line 428: "a _fixed_ collection" - the structure/size is fixed, but not the contents

Line 475: "sweeeps" -> "sweeps"

Ad 2.2.3 
This functionality already exists in perhaps a limited form - see the attached source file.
[access_func.f90.txt](https://github.com/j3-fortran/generics/files/6626225/access_func.f90.txt)

Ad 2.3
It will be quite important for this feature to generate clear error messages. It seems that at least early version of the C++ STL generated completely incomprehensible error messages if some mistake was made.

Possible additional use case:
I am not entirely sure this should be considered a use case for templates/generic programming, but several times in the not too distant past the possibility of associating units with variables and values has been discussed. In my field these units do not necessarily conform to the SI system of units: g carbon/m3 is then to be distinguished from g nitrogen/m3 for instance.

That will probably be equivalent to entering a dragon's cave, but we all know that dragons are keen collectors of baubles and gold ;).

tclune:
@arjenmarkus Thanks.  I'll try to integrate these into the version that is already on the J3 site  (will be 144r2).   I've already accumulated a few other typo fixes and new sentences.

tclune:
> Ad 2.1.2
> It is discussed later on, but it seems quite important to be able to define the requirements on the types that can be passed to the template: what operations are required, what relations there are between different arguments or template parameters and types. (How is that done in C++?)

It definitely is quite important to be able to do this.  (Though not all yet agree on that point.)     I've perhaps been too cautious in avoiding anything too concrete in this direction in the hopes of getting agreement on use cases that pretty much force the issue.

tclune:
> Ad 2.1.3
> What would happen in a MERGE algorithm if the two collections are NOT sorted? Is that a runtime error?

Presumably that or the result is truncated in some way.    I'm happy to ignore that for the moment as an implementation detail that is orthogonal to the template aspect.  (Mostly because this use case was contributed by someone else, and I don't know what they intended.)

tclune:
> Line 379: "is not recursive" - do we not consider recursive templates? They seem to provide quite some power in C++ (besides undoubtedly quite some complications ;))

At this time we do not have a compelling use case.    We can add one later if that arises.   I think so long as we have compile time conditionals that allow the recursion during instantiation to terminate that there is not a fundamental issue.    I just wanted to emphasize here that recursion is not involved in case someone things such power is dangerous/hard.

tclune:
> Ad 2.2.3
> This functionality already exists in perhaps a limited form - see the attached source file.
> [access_func.f90.txt](https://github.com/j3-fortran/generics/files/6626225/access_func.f90.txt)

I presume that you mean 2.2.6 (accessors)?       I agree that the recent extensions solve a good chunk of the concern, and was not certain whether they were allowed on the LHS of an assignment.  Good to know that they are.  The use case can still stand in case there are other aspects that cannot be so nicely  addressed.  Quite possibly this will not yield any requirements in the subsequent paper.

tclune:
> Ad 2.3
> It will be quite important for this feature to generate clear error messages. It seems that at least early version of the C++ STL generated completely incomprehensible error messages if some mistake was made.
> 

Yes - this is another case where I was perhaps being too subtle.  Malcolm has indicated that he thinks a good compiler could produce good error messages without such a feature.  (OTOH, he liked the fact that having this is easy on the compiler.)  I'll look again to see if I can add a sentence specifically about error messages.

> Possible additional use case:
> I am not entirely sure this should be considered a use case for templates/generic programming, but several times in the not too distant past the possibility of associating units with variables and values has been discussed. In my field these units do not necessarily conform to the SI system of units: g carbon/m3 is then to be distinguished from g nitrogen/m3 for instance.
> 
> That will probably be equivalent to entering a dragon's cave, but we all know that dragons are keen collectors of baubles and gold ;).

Unfortunately, it is possibly an excellent use case, but it would carry too much risk of killing the paper.  Lots of emotional baggage amongst the old timers on this one.    Do you think it would drive any requirements that don't stem from the existing use cases?   If so maybe we could put some proverbial lipstick on it and sell it under a different guise?

arjenmarkus:
I am aware of the emotional baggage :). And apart from that an attempt to capture the concept and what a programmer may expect is probably as lengthy as the paper is now. I merely wanted to bring it up, because, well, it is a topic that has arisen many times and seems akin to, if not a true example of, generic programming. I will have to think whether it poses additional requirements or capabilities.

PR #34, opened 2021-08-04
Initial draft.

tclune:


Requirements.md
```diff
@@ -3,3 +3,19 @@ from the various use cases.   Ideally this will drive the creation of a formal
...
+
+
+Terminology used in this paper:
+
+A template is a program unit that is parameterized by one or more
+template parameters.
+
+A template instantiation is the program unit defined by a template and
+a particular set of actual template parameters.
+
```
mleair, line 16: Based on today's meeting, it sounds like we might want full instantiation (e.g., use template T) and partial instantiation (e.g., use template T, only : x,y). If so, I suggest we expand the "template instantiation" definition to include full and partial instantiation.

PR #35, opened 2021-08-04
TKR encapsulation requirement

tclune:


tclune:
We should cite use cases.

I intentionally left out "attributes" in this initial draft, but this discussion should address this.

Requirements.md
```diff
@@ -3,3 +3,11 @@ from the various use cases.   Ideally this will drive the creation of a formal
 paper for generics that will be submitted to the Fortran committee.
 
 Each requirement should specify one or more motivating use cases.
+
+
+* The generics facility should provide a mechanism that enables
```
mleair, line 8: I suggest, "A template should provide a mechanism that enables..."

tclune: I've come up with a new formulation of this requirement.   And it is not really a property of the template, but the wrapper type, which might not be a template.

PR #36, opened 2021-08-04
Invalid code in conditional branches.

tclune:


tclune:
This feature is rather controversial.  But we can collect arguments for/against here.   More importantly, we should identify use cases that might necessitate this feature and consider variant approaches. (E.g., specialization)

tclune:
Really struggled with the wording on this one.

requirements.txt
```diff
@@ -0,0 +1,6 @@
+
+* Certain language constraints should be relaxed within conditional
+  branches that are not selected by actual template parameters for
+  constant conditional constructs.
+  
+
```
tclune, line 6: This wording is really more of a spec.  We want something that is less about the how and more about the what.

mleair: @tclune  Is this constant folding of template parameters? Can you give an example?

tclune: Not sure what "constant folding" is, so maybe?        Not sure I can quickly come up with a plausible case, but consider a template that has a logical template prarameter `is_writable`  which indicates that the type is either intrinsic or supported by DTIO.  One could then have
```f90
if (is_writable) then
   print*, x
end if
```
The trick her is that we would still want the code to be parsable, so it may be nigh impossible to list all the constraints that do not apply in such a context.   

Another example would be say a container of scalars where you wanted to allow deferred length strings, but not PDTs in general.   Here, variable declarations would need to be a bit different and one would actually want the "conditional" to be in the specification section of a module or procedure.   So much more like an `#ifdef`.

I think we all want a better approach.  And in the mean time this requirement needs to be stated more abstractly (somehow).

mleair: I think what you described is basically constant folding. It's when a preprocessor or compiler evaluates and simplifies expressions that are statically known (i.e., known at preprocessor-time and/or compile-time). Typically constant folding leads to other optimizations too such as dead code elimination, loop invariant removal, etc. But for our case, we might want to be able to selectively enable/disable code portions through template parameters. Constant-Folding can help with that. In fact, we may not need the "use only" for partial instantiation since this could also allow a way to partially instantiate a template through a parameter that selects certain code portions.

PR #37, opened 2021-08-04
Create an example for generic algorithms on a container

everythingfunctional:


scenarios/generic_algorithms/sparse_array_m.f90
```diff
@@ -0,0 +1,106 @@
+module sparse_array_m
+
+    concept :: copyable
```
tclune, line 3: I still think that one needs a dummy "argument" here to indicate the role of T.    It is "obvious" in this case with only one parameter, but once you have 2, you need to be able to select which type plays which role when invoking the concept name further on.

scenarios/generic_algorithms/sparse_array_m.f90
```diff
@@ -0,0 +1,106 @@
...
+        private
+        type(T), allocatable :: elements(:)
+        integer, allocatable :: indices(:)
+    contains
+        procedure :: insert_at
+        procedure :: get
+    end type
+contains
+    subroutine insert_at<T>(self, index, element)
+        requirements(T) :: copyable
```
tclune, line 18: As with above, it is "obvious" that T is the same as T above.  But with 2 parameters, it would not be.

scenarios/generic_algorithms/sparse_array_m.f90
```diff
@@ -0,0 +1,106 @@
+module sparse_array_m
```
tclune, line 1:
```suggestion
module sparse_array_m(T)
```

scenarios/generic_algorithms/sparse_array_m.f90
tclune:
```suggestion
    concept :: copyable(T)
```

scenarios/generic_algorithms/sparse_array_m.f90
tclune:
```suggestion
        requirements(T) :: copyable(T)
```

scenarios/generic_algorithms/sparse_array_m.f90
```diff
@@ -0,0 +1,106 @@
...
+                type(T), intent(in) :: x
+                type(U) :: y
+            end function
+        end interface
+        procedure(transformation_i) :: transformation
+        type(sparse_array_t<T>), intent(in) :: xs
+        type(sparse_array_t<U>) :: ys
+
+        ys%indices = xs%indices
+        ys%elements = etransformation(xs%elements)
```
tclune, line 49:
```suggestion
        ys%elements = transformation(xs%elements)
```

scenarios/generic_algorithms/sparse_array_m.f90
```diff
@@ -0,0 +1,106 @@
...
+    contains
+        elemental function etransformation(x) result(y)
+            type(T), intent(in) :: x
+            type(U) :: y
+
+            y = transformation(y)
```
tclune, line 55:
```suggestion
            y = transformation(x)
```

scenarios/generic_algorithms/sparse_array_m.f90
everythingfunctional: Part of me thinks for single type concepts (which I think is likely to be the majority) the "argument" should be optional, but I'm ok with being explicit for early examples.

scenarios/generic_algorithms/sparse_array_m.f90
```diff
@@ -1,6 +1,6 @@
-module sparse_array_m
+module sparse_array_m(T,U)
```
tclune, line 1: Oops - I realized just as I walked away last night that the entire purpose of this exercise was to demonstrate a case where 'U'  would only be a parameter to the "extra" procedures.    Definitely want to discuss further at meeting today.

```suggestion
module sparse_array_m(T)
```

tclune:
Single-type concepts certainly would simplify the possible applications.   How certain are you that such would cover your intended uses?      E.g., containers sure, but seems like most other templates I can contemplate involve relations among types.   But with a few simple one-operator filler like "printable".

everythingfunctional:
I'm sure single-type concepts won't cover all use cases, but it will cover a lot.

tclune:
@everythingfunctional could you provide some arguments as to why you think single-type concepts have a big footprint?    E.g., real use cases in other languages?    Arguments that they supplement/simplify the multi-type templates?          For the moment, I just cannot see past the example in either direction.

everythingfunctional:
Maybe it's just the simple examples, but I can think of a handful.

* Printable
* Stringable
* Copyable
* Equatable (e.g. T == T)
* Hashable
* Comparable (i.e. T > T)

Those would be sufficient for a lot of generic containers (i.e. Vector, Dictionary, Set)

tclune:
Yes - I agree that for containers this single-type feature works well.   But for any others?   Overkill for swap.   sparse linear algebra?      I suppose it might work well for many "intrinsic-like" procedures.

everythingfunctional:
Obviously there will be cases where having two (or more) type arguments to a concept is necessary, but I'm a bit partial to the capability that C++ enabled where something like

```C++
template <type T>
Stringable(T)
auto to_string(T object) {...}
```
can be simplified to something like

```C++
auto to_string(auto Stringable object) {...}
```

(and I'm sure I didn't get the syntax write)

But perhaps that's something that could be a future feature.

tclune (approved):
Thanks.  I'll do my best to have a parameterized modules analog to contrast with by the meeting on Tuesday.

PR #38, opened 2021-09-07
Variant sparse array implementation.

tclune:
This one only uses parameterized modules. I.e, does not use templated
types/procedures.

### DIFF
--- a/scenarios/generic_algorithms/alt_sparse_array_m.F90
+++ b/scenarios/generic_algorithms/alt_sparse_array_m.F90
@@ -166,7 +166,7 @@ program example_usage
     use sparse_array_m(integer), only: int_sparse_array => sparse_array_t
     use sparse_array_m(varying_string), only: string_sparse_array => sparse_array_t
 
-    use sparse_arary_map_m(integer, varying_string: map
+    use sparse_arary_map_m(integer, varying_string), only: map
     use sparse_arary_filter_m(varying_string), only: filter
     use sparse_arary_reduce_m(varying_string, ...), only: reduce
 
--- a/scenarios/generic_algorithms/alt_sparse_array_m.F90
+++ b/scenarios/generic_algorithms/alt_sparse_array_m.F90
@@ -1,0 +1,234 @@
+! This attempts to do the same thing as sparse_array_m.F90 but just using
+! parameterized modules.  I.e., without templated types/procedures.
+
+! First we have a parameterized module that is only about the
+! implementation of the sparse array itself.  This isolates the
+! container from the additional type parameters neeed by various
+! high-level operations (map, filter, ...)
+
+module sparse_array_m(T)
+   type :: T
+   end type T
+   
+   concept :: copyable(T)
+     assignment: T = T
+   end concept
+
+   type :: sparse_array_t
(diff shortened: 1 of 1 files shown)


scenarios/generic_algorithms/alt_sparse_array_m.F90
```diff
@@ -0,0 +1,234 @@
...
+
+end module
+
+
+program example_usage
+    use iso_varying_string, only: varying_string, assignment(=), operator(//), len
+    use sparse_array_m(integer), only: int_sparse_array => sparse_array_t
+    use sparse_array_m(varying_string), only: string_sparse_array => sparse_array_t
+
+    use sparse_arary_map_m(integer, varying_string: map
```
everythingfunctional, line 169:
```suggestion
    use sparse_arary_map_m(integer, varying_string), only: map
```

scenarios/generic_algorithms/alt_sparse_array_m.F90
```diff
@@ -0,0 +1,234 @@
...
+    use sparse_arary_filter_m(varying_string), only: filter
+    use sparse_arary_reduce_m(varying_string, ...), only: reduce
```
everythingfunctional, line 171:
```suggestion
    use sparse_arary_reduce_m(varying_string, varying_string), only: reduce
```

scenarios/generic_algorithms/alt_sparse_array_m.F90
```diff
@@ -0,0 +1,234 @@
...
+! parameterized modules.  I.e., without templated types/procedures.
+
+! First we have a parameterized module that is only about the
+! implementation of the sparse array itself.  This isolates the
+! container from the additional type parameters neeed by various
+! high-level operations (map, filter, ...)
+
+module sparse_array_m(T)
+   type :: T
+   end type T
```
everythingfunctional, line 11: Do you want this here? Doesn't that make the type that `sparse_array_t` stores different from whatever the template argument was?

tclune: No.  This is how you tell the compiler that the module parameter "T" is a type name as opposed to a LOGICAL, INTEGER, operator, ...   Just part of the syntax for parameterized modules.   Or I'm misunderstanding you.

everythingfunctional: But shouldn't the template work for intrinsic types and derived types? If you say T must be a type, are you saying that it can't be intrinsic?

tclune: No, that is not what this syntax means for a _module parameter_.     It means that T is a type name.  Nothing more.   (I've not looked closely at Van's paper, but I will be astonished at any other interpretation.)  Note it is also not saying anything about the data components nor methods of T.   

In our reboot, we could consider something more like   `type(TYPENAME) :: T`   

which introduces a new defined type "TYPENAME".   This might be somewhat clearer but at the cost of a new type.

everythingfunctional: ok. Didn't realize it was from Van's proposal.

PR #40, opened 2021-09-13
create an example to start introducing some category theory

everythingfunctional:
I want to start introducing some ideas to examine the possibilities of default implementations, and conditional compilation


tclune:
Nice.   Looking forward to filling in the error handling use case in this manner and floating it past other committee members.    Also very much want to see Magne's response on dealing with the conditional code issue here.

tclune (approved):
Accepting to get it into the repo.

PR #43, opened 2021-09-29
Multiple instantiations and PRIVATE

tclune:


everythingfunctional (approved):
Well formulated. I like it.

Requirements.md
```diff
@@ -3,3 +3,10 @@ from the various use cases.   Ideally this will drive the creation of a formal
 paper for generics that will be submitted to the Fortran committee.
 
 Each requirement should specify one or more motivating use cases.
+
+Requirement:   It must be possible to write a subroutine that has access to the private components of more than one instantiation of a templated type.    Without this requirement, multiple instantiations of a type would generally require the procedure to be defined in a separate module and would therefore not have access to the private components.
```
everythingfunctional, line 7:
```suggestion
Requirement:   It must be possible to write a procedure that has access to the private components of more than one instantiation of a templated type.    Without this requirement, multiple instantiations of a type would generally require the procedure to be defined in a separate module and would therefore not have access to the private components.

```

PR #44, opened 2021-09-29
Example to focus on PDT limitation

tclune:
- Type bound procedures play poorly with PDT.  An example is
  implemented 3 times: PDT, parameterized modules, and templated
  types. Several issues are worth discussion.

### DIFF
--- a/scenarios/pdt-dilemna/parameterized_modul.F90
+++ b/scenarios/pdt-dilemna/parameterized_modul.F90
@@ -1,0 +1,38 @@
+module foo_mod(k)
+   use, intrinsic :: iso_fortran_env
+   implicit none
+
+   integer :: k
+   
+   type :: foo
+      real(kind=k), allocatable :: data(:)
+   contains
+      procedure :: s
+   end type foo
+
+contains
+
+
+   function s(this)
+      class(foo<k>), intent(in) :: this
+      real(kind=k) :: s
+      s = sum(this%data)
+   end function s
+
+end module foo_mod
+
+program main
+   use foo_mod
+   use, intrinsic :: iso_fortran_env
+   implicit none
+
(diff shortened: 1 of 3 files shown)


scenarios/pdt-dilemna/parameterized_modul.F90
```diff
@@ -0,0 +1,38 @@
...
+
+end module foo_mod
+
+program main
+   use foo_mod
+   use, intrinsic :: iso_fortran_env
+   implicit none
+
+   type(foo<real(kind=REAL32)>) :: x32
+   type(foo<real(kind=REAL64>)) :: x64
```
everythingfunctional, line 30:
```suggestion
   use, intrinsic :: iso_fortran_env
   use foo_mod(real32), only: foo32 => foo
   use foo_mod(real64), only: foo64 => foo

   implicit none

   type(foo32) :: x32
   type(foo64) :: x64
```

I think this was probably just a copy past error, but that usage would look more like this.

scenarios/pdt-dilemna/template.F90
```diff
@@ -0,0 +1,39 @@
...
+   use, intrinsic :: iso_fortran_env
+   implicit none
+   
+   type :: foo<k>
+      integer :: k ! not a component!
+      
+      real(kind=k), allocatable :: data(:)
+   contains
+      procedure :: s
+   end type foo
```
everythingfunctional, line 11:
```suggestion
   concept :: is_integer(k)
      integer :: k
   end concept
   type :: foo<k>
      requires :: is_integer(k) ! not a component!
      
      real(kind=k), allocatable :: data(:)
   contains
      procedure :: s<k> ! or maybe the implementation should go here instead?
   end type foo
```

PR #46, opened 2021-11-05
An example using new TEMPLATE notation.

tclune:
The files in this directory are an attempt to demonstrate several
ideas that were bantered about at the last subgroup telecon using
the AXPY algorithm as an example.

  1. Introduce a new TEMPLATE construct - sort of like an inner
     parameterized module.
  2. How can a map procedure be combined with a rank-agnostic Array
     wrapper template, to provide operations on the wrapped elements?
  3. Use procedures instead of operators as template arguments.  (More
     general, and sidesteps some anoying notation issues to a degree.)

### DIFF
--- a/scenarios/axpy_with_arrays/AXPY.F90
+++ b/scenarios/axpy_with_arrays/AXPY.F90
@@ -1,0 +1,54 @@
+module AXPY_m
+
+   public :: axpy_t ! template
+   public :: scalar_op ! restriction
+   public :: assignable ! restriction
+   
+   RESTRICTION elemental_op(T, U, V, F)
+     elemental function F(x,y) result(z)
+       type(T), intent(in) :: x
+       type(U), intent(in) :: y
+       type(V) :: z
+     end function F
+   END RESTRICTION
+
+   RESTRICTION assignable(T, U)
+      elemental assignment(=)(x,y)
+      type(T), intent(out) :: x 
+      type(U), intent(in) :: y
+      end assignment
+   END RESTRICTION
+
+   template axpy_t(T_a, T_x, T_y, T_ax, T_axpy, times, plus)
+
+      type :: T_a
+      end type T_a
+      type :: T_x
+      end type T_x
+      type :: T_y
+      end type T_y
+      type :: T_ax
+      end type T_ax
+      type :: T_axpy
+      end type T_axpy
+
+      procedure :: times
+      procedure :: plus
+      
+      REQUIRES elemental_op(T_a, T_x, T_ax, times)
+      REQUIRES elemental_op(T_ax, T_y, T_axpy, plus)
+      REQUIRES assignable(T_y, T_axpy)
+
+   contains
+
+      subroutine AXPY(a, x, y)
+         type(T), intent(in) :: a
+         type(U), intent(in) :: x
+         type(V), intent(inout) :: y
+
+         y = plus(times(a,x), y)
+      end subroutine AXPY
+
+   END TEMPLATE
+
+end module AXPY_m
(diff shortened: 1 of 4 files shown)


tclune:
Note that this example also does not allow direct operations between "raw" intrinsics (scalar or array), but instead requires all operands to be wrapped with the generic Array wrapper.

scenarios/axpy_with_arrays/AXPY.F90
```diff
@@ -0,0 +1,54 @@
...
+      REQUIRES assignable(T_y, T_axpy)
+
+   contains
+
+      subroutine AXPY(a, x, y)
+         type(T), intent(in) :: a
+         type(U), intent(in) :: x
+         type(V), intent(inout) :: y
+
+         y = plus(times(a,x), y)
```
rouson, line 49: This is crying out for actual operator notation.   I have many times used lines like this to demonstrate the considerable improvement in clarity that operators provide.  The AXPY case isn't so bad, but one only has to add a few more terms to the expression for it to rapidly descend into being essentially indecipherable without translating it to operator syntax.

scenarios/axpy_with_arrays/README.md
```diff
@@ -0,0 +1,59 @@
...
+2. We have no way to specify that arrays are conformable.  There is a
+   runtime aspect of this that is inescapable, as the underlying
+   shapes can be different, but in theory I would like the `map_t`
+   template to require that either all ranks are equal except those
+   that are scalar:
+	 - `(rank_U == rank_T) .or. rank_u*rank_T = 0`
+     - `rank_V == max(rank_T, rank_U)`
+   Do we want to be able to do this?
+3. Procedures might be more general, but this example really does
+   scream for using operators.
```
rouson, line 51: Funny.  Sounds like we're on the same page.  My "crying out" is your "screaming."  LOL

scenarios/axpy_with_arrays/AXPY.F90
```diff
@@ -0,0 +1,54 @@
...
+   RESTRICTION assignable(T, U)
+      elemental assignment(=)(x,y)
+      type(T), intent(out) :: x 
+      type(U), intent(in) :: y
+      end assignment
+   END RESTRICTION
+
+   template axpy_t(T_a, T_x, T_y, T_ax, T_axpy, times, plus)
+
+      type :: T_a
```
rouson, line 24: double colons are optional without an attribute so you might remove them here and in many places below

rouson (approved):

scenarios/axpy_with_arrays/Array.F90
```diff
@@ -0,0 +1,70 @@
...
+      end type t
+      integer :: rank
+
+      type :: Array
+         type(T), rank(rank), allocatable :: elements
+      end type Array
+      
+      interface assignment(=)
+         module procedure assign_from_raw
+         module_procedure assign_to_raw
```
zoziha, line 16:
```suggestion
         module procedure assign_to_raw
```
May be a minor typo here :)

scenarios/axpy_with_arrays/AXPY.F90
tclune: Oh certainly.  But subgroup decided procedures were the priority.  I wanted to see how severe the consequences were ...   I really/unfortunately think we need both.

scenarios/axpy_with_arrays/AXPY.F90
tclune: But ... that's the way I roll ... :-)

scenarios/axpy_with_arrays/Array.F90
tclune:
```suggestion
         module procedure assign_to_raw
```
yup

PR #49, opened 2022-01-03
Draft requirements paper.

tclune:


certik:
Thanks for submitting it @tclune! My personal goal is to first implement at least some simpler cases in LFortran as a prototype (we are still working on it). Once we have something, I'll join your calls.

everythingfunctional:
One thought came to mind while reading the requirements. Where are templates and requirements allowed to be defined, and where can a template be instantiated? I'm inclined to say the declaration section of a program, module or subprogram. And now that I think about it more, perhaps that belongs in a Specs paper.

everythingfunctional:
The examples might suggest that restrictions are defined inside a template, and that is how they are imposed. But we'd like to be able to define restrictions outside a template, and then be able to impose restrictions on a template that are defined elsewhere. I think we should either modify the examples to demonstrate this, or explicitly state it somehow, possibly both.

J3-Papers/21-144r1.txt
```diff
@@ -0,0 +1,780 @@
...
+but in some cases these are also expected to drive different sets of
+requirements.  Use cases that have redundant requirements are
+generally suppressed in this paper, with the exception being when an
+otherwise redundant use case might resonate with segments of the user
+community or the committee.
+
+
+The identified categories of use cases are:
+
+    * Geniric Algorithms (2.1)
```
rouson, line 66: * Generic Algorithms (2.1)

J3-Papers/Generics_Requirements.txt
```diff
@@ -0,0 +1,139 @@
...
+           T + U => V
+        END RESTRICTION
+
+     
+  C. A template shall have one or more template parameters.
+
+  D. Each template parameter shall be one of the following:
+     - type spec
+     - constant expression
+     - generic procedure
```
rouson, line 68: Does "generic procedure" in this context refer to something new that we will be introduced in Fortran 202Y?  The Fortran 2018 standard already uses the terms "generic procedure name" and "generic intrinsic procedure" and even uses the term "generic procedure" in one place without "name" or "intrinsic."  We should probably find a way to differentiate what is referenced here from these various other terms in the standard.

rouson (changes requested):

J3-Papers/Generics_Requirements.txt
tclune: I meant the existing usage.   We allow the parameter to be the generic name - very different than for dummy arguments which must be specific procedures.   Happy to hear better wording to make this clear.

tclune:
@everythingfunctional I certainly did not mean for the examples to suggest that restrictions be defined inside templates.   Should make the change.

I'm wondering if the issue of "where" is the type of thing for the spec paper?

J3-Papers/Generics_Requirements.txt
rouson: @tclune I'm confused about how the existing usage of "generic procedure" fits in here.  In existing usage, a generic procedure name would be either

1. The name of a generic interface, 
2. The name of a generic binding, or
3. A generic intrinsic procedure.

Item 1 above designates a set of procedures, one of which can be matched with the procedure reference by the disambiguation rules, whereas I suspect what we want is a name that refers to just one procedure that is parameterized.

Item 2 above is explicitly precluded because we disallow type-bound procedures.

Item 3 would not seem to be of much use as it's defined by the language and can't be parameterized by the user.

tclune: These are _parameters_ to a template.  Suppose you wanted to pass SIN  to a generic solver?   Right now, you cannot do that because it is generic.   But the reasons that disallow this for dummy arguments do not apply for template arguments.   Disambiguation in templates can happen at usage.

Similarly, we often do not have access to specific names provided by other modules.   And again, we might want to use procedures in such as arguments to a template.

So I am actually meaning (1) here, and would appreciate edits that (a) use proper terminology and (b) a note explaining this intent, as I'm sure others can be thrown by this issue as well.

tclune: Thinking about it a bit more, I think we need _both_ options: specific name and generic name.

arjenmarkus:
Just a quick comment - I have not read the entire document yet, but while reading it I thought that a more or less "natural" way to specify the requirements for a type to be used in SORT or MERGE and perhaps other such generic procedure is something along these lines:

`
type :: comparable
     integer :: value
contains
    generic :: operator(<) => compare
    procedure :: compare => compare_int
    procedure :: sort => sort_int ! or generic :: sort -- this I realise is the weak part ;)
end type 
type(comparable) :: x(10), y(10)
y = comparable%sort( x )
`
So the type name could be used to connect the algorithm to the requirements, in this case the possibility to detect if one value is lower than another value.

arjenmarkus:
Oops, apologies for the very poor formatting :(

tclune:
@arjenmarkus Approaches along the lines you suggest would work for many things, and indeed lots of discussion in subgroup was in this direction.   Currently, our thinking is that the operator should be a separate template parameter.  One advantage of this is that one could then sort the same list in multiple ways - ascending or descending, by passing `<` or `>`.     

This is less crucial for sorting wher `<` reigns supreme in most implementations,  but in other contexts my "+" might be your "//" and we want the templates to have maximal generality.

(@rouson And here again, we need the operator, e.g., `<` to be generic, as the user may not have access to a specific procedures for "<".  I'm actually not 100% certain off the top of my head as to whether you can pass a generic `<` as an operator  actual argument in standard Fortran, but it seems it would run afoul of the same issues that prevent passing non-specific procedures as arguments.)

rouson:
@tclune I believe it's _not_ possible to pass an operator as an argument. In case it's useful, it is possible to have an operator in the `only` clause of a `use` statement.

arjenmarkus:
Yes, I realised later that defining the comparison as part of the type makes it awkward to use alternative comparisons. Though you could do it by implementing the comparison operator in a way that it uses a pointer to a function for the actual comparison (which feels rather clumsy).
Some further remarks though:

- Section 2.6.2 Accessors: such a facility as asked for is already available via pointer functions. I have attached an example.
- Variant 2,  Arbitrary rank contained objects: the same example is meant to store arrays of arbitrary length and even scalars. I developed it in the fun project "Mathematics of Arrays" (see the Fortran discourse thread on the subject or https://github.com/wyphan/moa-fortran).
[view_general.f90.txt](https://github.com/j3-fortran/generics/files/7833085/view_general.f90.txt)

tclune:
@rouson OK - but we still want to be able to pass operators (and generic names) as template parameters.

tclune:
> * Section 2.6.2 Accessors: such a facility as asked for is already available via pointer functions. I have attached an example.

I really now wish I'd not pushed that paper at the same time as the requirements paper.  It was just for completeness, as I'd missed committing it earlier.  But yes, the current standard appears to support the desired functionality.   Just makes it an easy requirement.  :-)


> * Variant 2,  Arbitrary rank contained objects: the same example is meant to store arrays of arbitrary length and even scalars. I developed it in the fun project "Mathematics of Arrays" (see the Fortran discourse thread on the subject or https://github.com/wyphan/moa-fortran).
>   [view_general.f90.txt](https://github.com/j3-fortran/generics/files/7833085/view_general.f90.txt)

Yes the hard part of this is the interface between client code and the container.  It is easy enough to make a template for say a Vector container that has elements that are scalars or allocatable arrays depending on some parameter.  But Fortran requires those arrays to then be inside a trivial wrapper.      This is not so bad itself, except insofar as this may force some overhead on the scalar case.

Suppose we have a template Vector that takes a template parameter T to implement a vector of (scalar) T.     We cannot readily re-use that template to contain elements that are (Fortran) arrays of T, without defining a wrapper type on the client side.    However, we can add an additional parameter N for the rank of the contained elements and have the template do the wrapping/unwrapping in the various operations.   Or we can define a separate Template for the contained-array case, which wraps the array for the user.   Not catastrophic, but annoying IMO given the conceptual similarity.

everythingfunctional (approved):

tclune:
@rouson Can you re-approve the PR? not going through because it thinks a change was requested.

rouson (approved):

rouson:
@tclune @everythingfunctional I just approved.

PR #54, opened 2022-01-18
What can templates contain?

tclune:
We missed an important requirement that templates actually have useful
contents.  This is a first attempt.


everythingfunctional (approved):
Don't know how we missed that. Too obvious I guess. Anyways, LGTM.

PR #55, opened 2022-01-18
Improved examples for requirements

tclune:
Added lines that used requirements and illegal lines that use
unsupported operations.

### DIFF
--- a/J3-Papers/Generics_Requirements.txt
+++ b/J3-Papers/Generics_Requirements.txt
@@ -170,6 +170,20 @@ subgroup:
            REQUIRES MAGMA(U, OPERATOR(+))
            REQUIRES MAGMA(V, OPERATOR(*))
         ...
+
+        CONTAINS
+
+           SUBROUTINE DO_SOMETHING(x, y)
+	      TYPE(U) :: x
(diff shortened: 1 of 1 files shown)


J3-Papers/Generics_Requirements.txt
```diff
@@ -170,6 +170,20 @@ subgroup:
...
+
+           SUBROUTINE DO_SOMETHING(x, y)
+	      TYPE(U) :: x
+	      TYPE(V) :: y
+
+              x = x + x*x ! legal
+	      y = y + y*y ! legal
+
+              x = x + y ! illegal - no supporting requirement
+	      x = 2*x   ! illegal integer multiply not in requirements
```
everythingfunctional, line 184: Should these lines be commented out? I'm somewhat of a mind that examples should compile and execute as written.

tclune: No strong opinion here.   The question is how it is usually done in examples.

The literalist in me says that commented out lines are, er, not illegal.  :-)

everythingfunctional: Ok, I'm good with being consistent. Thanks for checking.

PR #76, opened 2022-03-07
Add some examples of using the example templates

everythingfunctional:
Add some examples of how a couple of the templates could be used.


tclune (approved):

arjenmarkus:
I have a few minor comments on this:

- How do intrinsic types and kinds fit in? (Possibly even parametrised types)
- I see that interfaces are consistently defined, like:
`interface filter
     module procedure filter_
end interface`
   Is that a necessary part of the template specification?
- In the interfaces (map_) the type T and U are not imported. Is that deliberate or an oversight?
- Line 62 has typos: "ys%indeces" in stead of "ys%indices"
- Is the template parameter (template sparse_array_tmpl(T) for instance) automatically public? Or does that not matter?
- In swap_m, assignment of the type(T) variables is used, How to require that (intrinsic) operations or assignment (intrinsic or otherwise) are defined? What syntax to use?

These are detailed remarks and perhaps they should be put elsewhere but reading the code examples I began to wonder about these details.

PR #77, opened 2022-03-08
Add a Fortran version to comparison.md

certik:


everythingfunctional:
Would it make more sense to just copy the example into the document? Rather than inserting a link.

certik:
I think you are right, I added the code itself in 76d9ee7dd17170abfea2612e899992f45231f5c3.

everythingfunctional (approved):

certik:
Thanks!

PR #80, opened 2022-04-05
Modifications from meeting

everythingfunctional:
Modifications, suggestions, typo fixes, etc. from the meeting today.


everythingfunctional:
@tclune we kind of discussed a little bit about Section C and decided we ought to be clear about what constitutes identical for all types of actual template parameters. They could use some wordsmithing, but I think this at least gets the general idea.

tclune (approved):

PR #86, opened 2022-07-05
Example implementations of templates to emulate selected intrinsic procedures

everythingfunctional:


examples/intrinsics/restrictions_m.f90
```diff
@@ -0,0 +1,26 @@
...
+        interface
+            function equals(x, y)
+                type(T), intent(in) :: x, y
+                logical :: equals
+            end function
+        end interface
+    end restriction
+
+    restriction comparable(T, equals, less_than)
+        requires equatable(T, equals)
```
tclune, line 18: Usually comparable does not require equatable.   Indeed, most of C++ STL only requires "<" and can "derive `==` as   `.not. (a < b) .and. .not (b < a)`.

examples/intrinsics/findloc_m.f90
```diff
@@ -0,0 +1,141 @@
+module findloc_m
```
tclune, line 1: I had thought that `findloc()` might be able to cheat once you have an elemental equals.

After the checks one could simply do `findloc(equals(array, value))`

examples/intrinsics/restrictions_m.f90
```diff
@@ -0,0 +1,26 @@
...
+    implicit none
+    private
+    public :: equatable, comparable
+
+    restriction equatable(T, equals)
+        type :: T
+        end type
+        interface
+            function equals(x, y)
+                type(T), intent(in) :: x, y
```
tclune, line 11: Are we going to allow ELEMENTAL in the restriction.  Or if not that then maybe:
```suggestion
                type(T), intent(in) :: x(..), y(..)
```

examples/intrinsics/restrictions_m.f90
everythingfunctional: Indeed, one can derive all the other 4 comparison operators given just two of:
* `>`
* `>=`
* `<`
* `<=`
* `==`
* `/=`

The choice of which two to require and in what order is somewhat arbitrary, except that if one cares only about equality and not ordering, only one of `==` and `/=` is required.

examples/intrinsics/findloc_m.f90
everythingfunctional: I don't know why I didn't think of that. You're right. Let me see if I can make that work.

examples/intrinsics/restrictions_m.f90
everythingfunctional: Currently, a procedure dummy argument cannot require elemental, but can require pure, and then an internal elemental procedure can be used. I'd be in favor of allowing elemental for procedure template parameters, but I'd be interested to hear other opinions.

examples/intrinsics/restrictions_m.f90
tclune: Maybe!    I've not seen anything in C++ STL that automates it via a template, but maybe was not looking.     OTOH, it may only be well-defined for iterators, which then are already a single template (I think).   E.g., STL C++ does not define a "<" for map (dictionary) itself, only "==" and "/=".

Thinking out loud as I'm interested ...
```
TEMPLATE RELATIONAL_OPERATORS(T, less)
    TYPE :: T; END TYPE
     INTERFACE
         PURE LOGICAL FUNCTION less(x,y)
              TYPE(T), INTENT(IN) :: x, y
         END FUNCTION LESS
     END INTERFACE         
     PRIVATE

     PUBLIC :: OPERATOR(<)     
     PUBLIC :: OPERATOR(<=)     
     PUBLIC :: OPERATOR(>)     
     PUBLIC :: OPERATOR(>=)     
     PUBLIC :: OPERATOR(==)     
     PUBLIC :: OPERATOR(==)     

     INTERFACE OPERATOR(<)
         procedure less
     END INTERFACE

     INTERFACE OPERATOR(>)
        procedure greater
     END INTERFACE

...
```
Too bad that you cannot then use the operators as arguments to other templates.    Need to expose the actual names as well.  :-(

everythingfunctional: Yep, I created a `comparisons_t` template. It works well I think.

examples/intrinsics/restrictions_m.f90
everythingfunctional: I utilized the technique above to provide elemental comparison operators given a `pure` less_than procedure

examples/intrinsics/restrictions_m.f90
everythingfunctional: I don't think you need to expose the procedures. No template will ever need more than the first one. Any template that wants to use the comparison operators will already need to require the type be comparable, so will have a `requires` statement and take the `less` procedure as a parameter. Then it can instantiate the operators itself and use them. Check out how I did it now.

tclune: I'm not so sure.   Consider where I've got an array that I want to sort and the sort template takes a procedure for ordering.    My type only has a less, so I've got to define a "greater" to sort the other way.     Your template gets me half way and then I've got to create a wrapper.  :-(

tclune:
Another thought ala Magne's use of sub-templates for optional arguments.   Imagine a GeneralizedIntrinsics template that takes a type T.    It then has subtemplates for things that need `equals` or `less`, etc.  Not sure how fruitful it would be, but might be a good tutorial case to show the power?

examples/intrinsics/restrictions_m.f90
everythingfunctional: Hmm, perhaps. I'll think about it. Might depend on how your sort template is defined.

PR #87, opened 2022-07-05
Updated existing swap example.

tclune:
- included support for array arguments
 - included support for pointer and allocatable arguments
 - included alternate implementation which has extra parameter for
   rank


everythingfunctional (approved):
LGTM

PR #89, opened 2022-07-11
Feature/update matmul example

tclune:
Made the example a bit more advanced.     I think I may have been a bit fast-and-loose with reusing names (matmul, sum, etc.)  OTOH, it may well be that I've renamed things that did not need renaming.   Goal should be to maximize legibility.


tclune:
I'll try to add some comments.   Even if they won't really explain the example to a casual reader, it will at least give me some prompts when I present ...

PR #93, opened 2022-07-19
fix typo

brandongc:
missing close parenthesis


zoziha:
In addition to this, there is a typo in the following line, `!=` should be `/=`:

https://github.com/j3-fortran/generics/blob/69e5cf6071ea951bb1ffdb21e5889c6fbd44e854/examples/vector/vector.F90#L42

zoziha (approved):

PR #96, opened 2022-10-06
add numeric requirements from Magne

everythingfunctional:
A hierarchical collection of requirements for numeric operations.


everythingfunctional:
I'd like to go back and forth with Magne a bit before accepting these, so hold off on merging for now.

tclune (approved):
Your matrix goes a bit further than mine; in particular you derive `SUM()` from `plus()` and `zero`, so maybe this should be the one used in the tutorial?  Probably not as simpler is probably better for the tutorial, but we certainly don't want to do both.

PR #97, opened 2022-10-17
Initial attempt at rich Linear Algebra example

tclune:
Need further discussion with Magne regarding how to nest templates.


examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+     type, deferred :: T
+     pure function op(x,y) result(z)
+        type(T), intent(in) :: x, y
+        type(T)) :: z
+     end function op
+   end requirement oper
+
+   requirement elemental_oper(T, op)
+     type, deferred :: T
+     pure function op(x,y) result(z)
```
everythingfunctional, line 16: I'm guessing you meant this?

```suggestion
     elemental function op(x,y) result(z)
```

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+
+
+   template Matrix_tmpl(T,plus,times,n)
+      requires elemental_oper(T,plus)
+      requires oper(T,times)
+      privawe
+
+      public :: Matrix
+      public :: operator(+)
+      public :: operator(*)
```
everythingfunctional, line 31: You don't have an `operator(*)` in this template. Did you mean `operator(.matmul.)`?

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+         type(T) :: elements(n,n)
+      end type
+
+      interface operator(+)
+         procedure :: plus_matrix
+      end interface operator(+)
+
+      interface operator(.matmul.)
+         procedure :: matmul_matrix
+      end interface operator(*)
```
everythingfunctional, line 43: The names don't match here. Which one did you want?

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+
+         interface zero
+            procedure :: zero_matrix
+         end interface zero
+
+      contains
+
+         pure function zero_matrix()
+            type(Matrix) :: zero_matrix
+            zero_matrix%elements = zero()
```
everythingfunctional, line 67: I think you meant

```suggestion
            zero_matrix%elements = zero_t()
```

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+     type, deferred :: T
+     pure function op(x,y) result(z)
+        type(T), intent(in) :: x, y
+        type(T)) :: z
+     end function op
+   end requirement elemental_oper
+
+
+
+   template Matrix_tmpl(T,plus,times,n)
```
everythingfunctional, line 24: `n` isn't declared.

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+
+      !======================================
+      template MatrixZero_tmpl(zero_t)
+      !--------------------------------------
+         requires elemental_oper(zero_t)
+         private
+         public :: zero
+
+         interface
+            elemental function zero_t()
```
everythingfunctional, line 54: Technically, I don't think this needs to be elemental.

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+         interface one
+            procedure :: identity_matrix
+         end interface one
+
+      contains
+
+         pure function identity_matrix()
+            type(Matrix) :: matrix_one
+
+            matix_one%elements = 0
```
everythingfunctional, line 97: I think this needs to be

```suggestion
            matix_one%elements = zero_t()
```

examples/linear_algebra/Matrix.F90
```diff
@@ -0,0 +1,274 @@
...
+     end template
+     !=====================================
+
+
+     !=======================================
+     template MatrixSubtraction_tmpl(minus_t)
+     !---------------------------------------
+        requires elemental_oper(T, minus_t)
+        private
+        public :: minus_t
```
everythingfunctional, line 111: I think you meant

```suggestion
        public :: minus
```

examples/linear_algebra/Matrix.F90
tclune: Yes - originally just had one requirement and then decided to split and updated the wrong one.

examples/linear_algebra/Matrix.F90
tclune: Not sure what you mean.   These do not need to match, but my intent was `OPERATOR(*)` in any event.

PR #99, opened 2022-10-23
Force implicit mapping

tclune:


tclune:
@everythingfunctional I should have deleted these PRs.    We did some changes in this direction separately in subgroup, and I think we'll just end up with a mess unless we do this collectively.

PR #100, opened 2022-10-23
Attempt at restricted INTERFACE in template.

tclune:


tclune:
UTI:

What if the  procedure parameters in a REQUIREMENT, themselves take a procedure.  E.g.,

```f90
REQUIREMENT  R(T, F)
      TYPE, DEFERRED :: T
      FUNCTION F(x, g) RESULT(y)
           INTERFACE 
                 FUNCTION g(x,y) RESULT(z)
                       TYPE(T) :: x, y, z
                 END FUNCTION
           END INTERFACE
       END INTERFACE
END REQUIREMENT

everythingfunctional:
I think what we want here is for `<template-interface-block>` to use it's own `<template-interface-body>`, where we say that the body has host association and no implicit mapping. Maybe we can (re)use what we've already started for restriction blocks?

I think we might also want `<template-procedure-definition>` (or whatever the existing equivalent is), so that we can say that they can use `<template-interface-block>`, but not regular `<interface-block>`.

For your UTI:

> What if the procedure parameters in a REQUIREMENT, themselves take a procedure. E.g.,

How can one do this in the existing interface blocks? I'm not sure I've tried it yet. Would you use an abstract interface block outside, bring in the abstract interface name and use it in a procedure statement? I think we can just say that you can specify interfaces for procedures that aren't in the requirement argument list, and that they are then abstract interfaces and can be used in procedure statements inside the other interfaces.

PR #104, opened 2023-03-22
Update examples

everythingfunctional:
Update many of our examples to align with current syntax and semantics.


certik:
Thanks @everythingfunctional !

@ansharlubis, some new tests that we can try for the generics in LFortran. :)

tclune (approved):

PR #106, opened 2023-05-24
Trying to get a bit done between telecons.

tclune:
1. Suggested terminology to replace "ultimate specification"
2. Attempt at 2 more "lattice" constraints.

### DIFF
--- a/J3-Papers/new_syntax_paper.txt
+++ b/J3-Papers/new_syntax_paper.txt
@@ -302,16 +302,14 @@ R2 then:
 
 *** UNREVIEWED TEXT (TOM) ***
 
-Constraint: If 2 terminal specifications of a <deferred-proc> have
-            characterictics C1 and C2, then their active specification
-            has characteristics C1.  Otherwise their active
-            specification is inconsistent.
+If 2 terminal specifications of a <deferred-proc> have characterictics
+C1 and C2, then their active specification has characteristics C1.
+Otherwise their active specification is inconsistent.
 
-Constraint: If 2 terminal specifications of a <deferred-proc> have
-            argument names A = {a_1, a_2, ..., a_na} and B = {b_1,
-            b_2, ..., b_nb}, then their active specification has names
-            A if na = nb, and a_i = b_i for i = 1, 2, ... na.
-            Otherwise their active specification is inconsistent.
+If 2 terminal specifications of a <deferred-proc> have argument names
+A = {a_1, a_2, ..., a_na} and B = {b_1, b_2, ..., b_nb}, then their
+active specification has names A if na = nb, and a_i = b_i for i = 1,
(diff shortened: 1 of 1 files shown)


J3-Papers/new_syntax_paper.txt
```diff
@@ -300,6 +300,21 @@ R2 then:
...
+Constraint: If 2 terminal specifications of a <deferred-proc> have
+            characterictics C1 and C2, then their active specification
+            has characteristics C1.  Otherwise their active
+            specification is inconsistent.
+
+Constraint: If 2 terminal specifications of a <deferred-proc> have
+            argument names A = {a_1, a_2, ..., a_na} and B = {b_1,
+            b_2, ..., b_nb}, then their active specification has names
+            A if na = nb, and a_i = b_i for i = 1, 2, ... na.
+            Otherwise their active specification is inconsistent.
```
everythingfunctional, line 314: This looks like the "all must interfaces must be spelled the same" option in the straw vote. Is that what we want to start with?

J3-Papers/new_syntax_paper.txt
```diff
@@ -300,6 +300,21 @@ R2 then:
   - if both R1 and R2 are deferred rank, then the active specification
     has deferred rank.
 
+*** UNREVIEWED TEXT (TOM) ***
+
+Constraint: If 2 terminal specifications of a <deferred-proc> have
+            characterictics C1 and C2, then their active specification
+            has characteristics C1.  Otherwise their active
```
everythingfunctional, line 307: Still need to say something about C1 and C2 being the same.

J3-Papers/new_syntax_paper.txt
tclune: No - but I figured this was a good starting point to get something in place.   It was the easiest one to state.   

If explicit overrides implicit then it will be a bit more ungainly to state.

J3-Papers/new_syntax_paper.txt
tclune: Oops typo.  That's why we do reviews ...

J3-Papers/new_syntax_paper.txt
```diff
@@ -300,6 +300,19 @@ R2 then:
   - if both R1 and R2 are deferred rank, then the active specification
     has deferred rank.
 
+*** UNREVIEWED TEXT (TOM) ***
+
+If 2 terminal specifications of a <deferred-proc> have characterictics
+C1 and C2, then their active specification has characteristics C1.
+Otherwise their active specification is inconsistent.
```
tclune, line 307:
```suggestion
If 2 terminal specifications of a <deferred-proc> have characterictics
C1 and C2, then their active specification has characteristics C1 if C1 and C2 are the
same.  Otherwise their active specification is inconsistent.
```

PR #113, opened 2023-09-24
Draft position paper on exporting intrinsic operators.

tclune:


J3-Papers/export-generic.txt
```diff
@@ -0,0 +1,98 @@
...
+
+(3) Do nothing.  Existing language constraints already prevent the
+    instantiation in the example above.    Users may find somewhat
+    obscure compilation error messages.
+
+Subgroup found that (1) was too severe.  We have numerous examples
+that potentially benefit from exporting intrinsic operators extended for
+derived types defined within the same template.
+
+Subgroup preferred option option (2) over (3), primarily on the basis
```
everythingfunctional, line 82:
```suggestion
Subgroup preferred option (2) over (3), primarily on the basis
```

J3-Papers/export-generic.txt
```diff
@@ -0,0 +1,98 @@
...
+
+4. Conclusion
+=============
+
+Subgroup proposes to add a constraint to the syntax in 15.4.3.2
+Interface block along the following lines:
+
+Constraint: If <generic-spec> is <extended-intrinsic-op>, then at
+            least one dummy argument in each <interface-specification>
+            shall be non-deferred and non-intrinsic.
```
everythingfunctional, line 96: I think the intent is correct here, but I think there's some nuances that need to be wordsmithed.

tclune: Yes - I stared at the BNF a bit, and found that this will be a bit difficult to specify succinctly.   I don't think this paper needs to get it "right", but just indicative.  To that end maybe not even make it look like a constraint:

"... a constraint along the lines that disallows public intrinsic operators in a template unless all of the specific procedures involve at least one argument that is neither intrinsic nor deferred."

PR #118, opened 2024-05-14
Misc edits for implied-shape array, RANK clause, and glossary

aury6623:
Implementing the UTIs we put in during the last meeting. Also a couple other changes that are relevant.

Note my note!
```
Note: The change to C864 above, along with the normative text change,
      also allow RANK to work with a PARAMETER constant, not only a
      deferred constant, because they say "named constant" instead
      of "deferred constant". I think that's fine!
```
(I didn't modify the change to C864 in this PR, it's right above one of the changed portions.)

### DIFF
--- a/J3-Papers/misc-template-edits.txt
+++ b/J3-Papers/misc-template-edits.txt
@@ -16,10 +16,15 @@ been identified so far.
 2. Edits required for deferred arguments
 ========================================
 
-* Insert new glossary term after 3.38.2
+* Insert new glossary term 3.34.x
 
-  [named deferred constant]
+  [deferred constant]
   named data object with the CONSTANT attribute (????)
+
+* Modify glossary term 3.34.2 "named constant" to
+
+  [named constant]
+  named data object (3.42) with the PARAMETER attribute, or a deferred constant
 
 * 7.3.2.1 Type specifier syntax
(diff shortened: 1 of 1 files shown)


J3-Papers/misc-template-edits.txt
```diff
@@ -81,6 +112,27 @@ would make constraint for deferred constants clearer.
...
+    the ALLOCATABLE or POINTER attribute, it specifies that it is a
+    deferred-shape array; otherwise, if it is a named constant, it
+    specifies that it is an implied-shape array with all lower
+    bounds equal to one; otherwise, it specifies that it is an
+    assumed-shape array with all the lower bounds equal to one.
+
+Note: The change to C864 above, along with the normative text change,
+      also allow RANK to work with a PARAMETER constant, not only a
+      deferred constant, because they say "named constant" instead
+      of "deferred constant". I think that's fine!
```
everythingfunctional, line 134:
```suggestion
Note: The change to C864 above, along with the normative text change,
      also allow RANK to work with a PARAMETER constant, not only a
      deferred constant, because they say "named constant" instead
      of "deferred constant". I think that's fine!
```

everythingfunctional (approved):

PR #119, opened 2024-05-14
Issue with 'innermost scoping unit' constraint

aury6623:
There are a few issues with the constraints removed in this edit:

- I think that "innermost scoping unit" here is a bit too vague, especially because...
- This paper does not define \<deferred-arg-list\>, since that comes from papers 2 & 3. There should at least be a note about that.
- I think all of these need to constrain themselves to \<deferred-arg\>s appearing in their declaration statements, like the constraint for deferred procedures. For example, a \<deferred-const\> can appear in an expression context without it appearing in the \<deferred-arg-list\> of the innermost scoping unit, if it's host associated. It's only when the \<deferred-arg\> appears in a declaration context (including a REQUIRES statement) that it must be in the \<deferred-arg-list\>.

Given that, this PR contains my suggested new wording. I think we can combine all of these into a single constraint, plus a note.

The new constraint wording is loosely based on a similar constraint from the F2023 standard:

> C746 A \<type-param-name\> in a \<type-param-def-stmt\> in a \<derived-type-def\> shall be one of the \<type-param-name\>s in the \<derived-type-stmt\> of that \<derived-type-def\>.

But I couldn't find any perfectly analogous existing constraint that I could directly borrow wording from.

If we choose not to combine all 4 of these constraints into 1, we should at least combine the last two, as: "Each \<deferred-type\> or \<deferred-class\> shall appear...", since we do that in the constraint right above those which deals with the EXTENDS attribute.

### DIFF
--- a/J3-Papers/syntax-intro-and-deferred-args.txt
+++ b/J3-Papers/syntax-intro-and-deferred-args.txt
@@ -183,6 +183,15 @@ Note: Deferred arguments are local identifiers and are not externally
                              <<or>> <deferred-proc-declaration-stmt>
                              <<or>> <deferred-type-declaration-stmt>
 
+Constraint: A <deferred-arg> that appears in a <deferred-arg-decl-stmt>
+            shall be in the <deferred-arg-list> of the innermost
+            TEMPLATE, REQUIREMENT, or simple template procedure in
+            which it appears.
+
+Note: <deferred-arg-list> is part of the TEMPLATE, REQUIREMENT, and
+      STP constructs, and is defined in the 2nd and 3rd generics
+      formal syntax papers.
+
 3.1.1 Syntax for deferred constants
 
 <deferred-const-declaration-stmt> <<is>>
@@ -223,9 +232,6 @@ Constraint: If <explicit-shape-bounds-spec> appears in
 
 <deferred-const> <<is>> <name>
 
-Constraint: Each <deferred-const> shall appear in <deferred-arg-list>
-            of the innermost scoping unit.
-
 A <deferred-const> is a deferred constant.
 
 Some examples of declaring deferred constants are as follows.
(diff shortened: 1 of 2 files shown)


J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -183,6 +183,15 @@ Note: Deferred arguments are local identifiers and are not externally
...
 
+Constraint: A <deferred-arg> that appears in a <deferred-arg-decl-stmt>
+            shall be in the <deferred-arg-list> of the innermost
+            TEMPLATE, REQUIREMENT, or simple template procedure in
+            which it appears.
+
+Note: <deferred-arg-list> is part of the TEMPLATE, REQUIREMENT, and
+      STP constructs, and is defined in the 2nd and 3rd generics
+      formal syntax papers.
+
```
tclune, line 194:
```suggestion
Constraint: A <deferred-arg> that appears in a <deferred-arg-decl-stmt>
            shall be included in the <deferred-arg-list> of the 
            TEMPLATE, REQUIREMENT, or simple template procedure in
            which it appears.

Note: <deferred-arg-list> is part of the TEMPLATE, REQUIREMENT, and
      STP constructs, and is defined in the 2nd and 3rd generics
      formal syntax papers.

```

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -183,6 +183,15 @@ Note: Deferred arguments are local identifiers and are not externally
                              <<or>> <deferred-proc-declaration-stmt>
                              <<or>> <deferred-type-declaration-stmt>
 
+Constraint: A <deferred-arg> that appears in a <deferred-arg-decl-stmt>
```
everythingfunctional, line 186:
```suggestion
Constraint: A <deferred-arg> that appears in a <deferred-arg-explicit-stmt>
```

everythingfunctional (approved):

PR #120, opened 2024-05-14
Use ABSTRACT keyword to indicate an abstract deferred type, in order to resolve issues with the distinction of deferred-type vs deferred-class

aury6623:
There are inconsistencies with the usage of "deferred type"/\<deferred-type\> and "deferred class"/\<deferred-class\>

* https://github.com/j3-fortran/generics/blob/cc42f87f976b4a151111c349498e826ff5b8aa18/J3-Papers/syntax-intro-and-deferred-args.txt#L112-L114
* https://github.com/j3-fortran/generics/blob/cc42f87f976b4a151111c349498e826ff5b8aa18/J3-Papers/syntax-intro-and-deferred-args.txt#L318-L319
* https://github.com/j3-fortran/generics/blob/cc42f87f976b4a151111c349498e826ff5b8aa18/J3-Papers/syntax-intro-and-deferred-args.txt#L437-L449

In the third of those 3 snippets, we assume that "deferred type" and "deferred class" are mutually exclusive categories. However, in the first snippet, as well as the section headers in the document, we refer to both things together as "deferred type". We also say "can appear in a \<type-spec\>" in the first snippet, which is not true of deferred class. We also group them in the BNF, such as "\<deferred-type-declaration-stmt\>" which is either TYPE, DEFERRED or CLASS, DEFERRED. We need to be more precise with the terms here.

In the existing standard, "class" is not a term used to describe a type, and it's not really even a word used to describe a data object. I think the closest it gets to that is the "CLASS IS" syntax in a \<type-guard-stmt\>. CLASS is only a type specifier which declares polymorphic data objects. I think our problem might be the use of the word "class" at all to describe the concept we want; when I see CLASS in Fortran, I think "declaration of an object", not "definition of a type".

The `TYPE, DEFERRED :: type-name` declaration is supposed to be analogous to the `TYPE :: type-name` construct, right? So I think it would be more clear to use `TYPE, ABSTRACT, DEFERRED :: type-name` as the declaration, analogous to `TYPE, ABSTRACT :: type-name`.

With that notion, this PR is my proposal to address the inconsistencies between \<deferred-type\> and \<deferred-class\>.

If the `TYPE, ABSTRACT, DEFERRED :: type-name` syntax is not preferred, we could still revert back to the `CLASS, DEFERRED :: type-name` syntax and keep everything else in this edit the same, except the normative text would change to:

> If **[[ABSTRACT appears -> CLASS is used]]** in \<deferred-type-declaration-stmt\>, the declared
deferred types act as if they were abstract extensible derived types.
Otherwise, they act as if they were non-extensible types.

### DIFF
--- a/J3-Papers/misc-template-edits.txt
+++ b/J3-Papers/misc-template-edits.txt
@@ -26,13 +26,6 @@ been identified so far.
   Extend R702 <type-spec>:
   " <<or>> <deferred-type>"
 
-  Extend C703 from:
-    (R702) The <derived-type-spec> shall not specify an abstract
-    type (7.5.7)
-  To:
-    (R702) The <derived-type-spec> or <deferred-type> shall not specify
-    an abstract type (7.5.7) except when used as an <instantiation-arg>.
-
   Extend R703 <declaration-type-spec>:
   "TYPE(<deferred-type>)"
 
--- a/J3-Papers/misc-template-edits.txt
+++ b/J3-Papers/misc-template-edits.txt
@@ -26,11 +26,18 @@ been identified so far.
   Extend R702 <type-spec>:
   " <<or>> <deferred-type>"
 
+  Extend C703 from:
+    (R702) The <derived-type-spec> shall not specify an abstract
+    type (7.5.7)
+  To:
+    (R702) The <derived-type-spec> or <deferred-type> shall not specify
+    an abstract type (7.5.7) except when used as an <instantiation-arg> (????)
+
   Extend R703 <declaration-type-spec>:
   "TYPE(<deferred-type>)"
 
   Extend R703 <declaration-type-spec>:
-  "CLASS(<deferred-class>)"
+  "CLASS(<deferred-type>)"
 
 * 7.5.2.1 Syntax of a derived-type definition
 
@@ -104,7 +111,7 @@ would make constraint for deferred constants clearer.
   R1156 <type-guard-stmt>
       <<is>> TYPE IS ( <type-spec> ) [ <select-construct-name> ]
       <<or>> CLASS IS ( <derived-type-spec> ) [ <select-construct-name> ]
-      <<or>> CLASS IS ( <deferred-class> ) [ <select-construct-name> ]
+      <<or>> CLASS IS ( <deferred-type> ) [ <select-construct-name> ]
       <<or>> CLASS DEFAULT [ <select-construct-name> ]
 
 * 15.5.1 Syntax of a procedure reference
--- a/J3-Papers/misc-template-edits.txt
+++ b/J3-Papers/misc-template-edits.txt
@@ -31,7 +31,7 @@ been identified so far.
     type (7.5.7)
   To:
     (R702) The <derived-type-spec> or <deferred-type> shall not specify
-    an abstract type (7.5.7) except when used as an <instantiation-arg> (????)
+    an abstract type (7.5.7) except when used as an <instantiation-arg>.
 
   Extend R703 <declaration-type-spec>:
   "TYPE(<deferred-type>)"
--- a/J3-Papers/misc-template-edits.txt
+++ b/J3-Papers/misc-template-edits.txt
@@ -26,6 +26,13 @@ been identified so far.
   Extend R702 <type-spec>:
   " <<or>> <deferred-type>"
 
+  Extend C703 from:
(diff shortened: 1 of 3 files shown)


J3-Papers/misc-template-edits.txt
```diff
@@ -26,11 +26,18 @@ been identified so far.
   Extend R702 <type-spec>:
   " <<or>> <deferred-type>"
 
+  Extend C703 from:
+    (R702) The <derived-type-spec> shall not specify an abstract
+    type (7.5.7)
+  To:
+    (R702) The <derived-type-spec> or <deferred-type> shall not specify
+    an abstract type (7.5.7) except when used as an <instantiation-arg> (????)
```
aury6623, line 34: This last part, "except when used as an <instantiation-arg>", is needed regardless of the changes in this edit. We use \<type-spec\> for deferred types, but as-is that doesn't work because \<type-spec\>s need to be non-abstract, and we need to allow them to be abstract.
If modifying the constraint in this way is not acceptable, then we'll need to use something other than \<type-spec\>.

J3-Papers/syntax-template-and-instantiate.txt
```diff
@@ -389,25 +389,23 @@ Note: The previous two constraints constitute what is referred to as
...
-            <deferred-type>, a variable of that type shall be
-            permitted in a variable definition context.
+Constraint: An <instantiation-arg> that corresponds to an abstract
+            deferred type shall be an extensible derived type.
 
-Constraint: If an <instantiation-arg> is a <type-spec>, it shall not
-            specify a type that has a coarray potential subobject
-            component.
+Constraint: An <instantiation-arg> that corresponds to a deferred
+            type shall be permitted in a variable definition context.
```
aury6623, line 401: Actually I think this should be "...that corresponds to a **non-abstract** deferred type..." ?

J3-Papers/misc-template-edits.txt
tclune, line 34:
```suggestion
  Extend C703 from:
    (R702) The <derived-type-spec> shall not specify an abstract
    type (7.5.7)
  To:
    (R702) The <derived-type-spec> or <deferred-type> shall not specify
    an abstract type (7.5.7) except when used as an <instantiation-arg>.
```

J3-Papers/misc-template-edits.txt
```diff
@@ -26,11 +26,18 @@ been identified so far.
   Extend R702 <type-spec>:
   " <<or>> <deferred-type>"
 
+  Extend C703 from:
+    (R702) The <derived-type-spec> shall not specify an abstract
+    type (7.5.7)
+  To:
+    (R702) The <derived-type-spec> or <deferred-type> shall not specify
+    an abstract type (7.5.7) except when used as an <instantiation-arg>.
+
```
tclune, line 35:
```suggestion
 ```

Comment (aury6623) [inline J3-Papers/misc-template-edits.txt:35]: I put this back because I realized why it's necessary for this edit.

The "except when used as an \<instantiation-arg\>" is not directly relevant to this edit, but is important, so I gave it its own PR: https://github.com/j3-fortran/generics/pull/127

But indicating that \<deferred-type\> must not be abstract is important because it solves the issue @everythingfunctional was talking about: it prevents you from using an abstract deferred-type in a TYPE IS statement.

File: J3-Papers/misc-template-edits.txt [hunk 4]
```diff
@@ -104,9 +127,12 @@ would make constraint for deferred constants clearer.
   R1156 <type-guard-stmt>
       <<is>> TYPE IS ( <type-spec> ) [ <select-construct-name> ]
       <<or>> CLASS IS ( <derived-type-spec> ) [ <select-construct-name> ]
-      <<or>> CLASS IS ( <deferred-class> ) [ <select-construct-name> ]
+      <<or>> CLASS IS ( <deferred-type> ) [ <select-construct-name> ]
       <<or>> CLASS DEFAULT [ <select-construct-name> ]
 
+  Add constraint, or amend C1166 to include:
+    (R1156) <deferred-type> shall specify an extensible type.
+
```
aury6623, line 135: Note that "extensible type" here is a generalization of "abstract type" and allows us to expand to something like `TYPE, EXTENSIBLE :: T` in the future without needing to change this constraint. Abstract should imply extensible.

everythingfunctional:
Addressed by #131

PR #131, opened 2024-05-23
Use attributes for deferred type

everythingfunctional:
Uses attributes on `type, deferred ::` to enable use of polymorphism instead of `class, deferred ::`

### DIFF
--- a/J3-Papers/syntax-intro-and-deferred-args.txt
+++ b/J3-Papers/syntax-intro-and-deferred-args.txt
@@ -314,7 +314,7 @@ Constraint: A <deferred-type> entity shall not appear as
 Constraint: Each <deferred-type> shall appear in <deferred-arg-list>
             of the innermost scoping unit.
 
-A <deferred-type> is a type that is a deferred argument.
+A <deferred-type> is a deferred type.
 
 If ABSTRACT appears in <deferred-type-declaration-stmt>, the declared
 deferred types act as if they were abstract extensible derived types.
--- a/J3-Papers/syntax-intro-and-deferred-args.txt
+++ b/J3-Papers/syntax-intro-and-deferred-args.txt
@@ -109,15 +109,19 @@ REQUIRES and INSTANTIATE statements.
 A deferred constant is a deferred argument that can appear in constant
 expressions within a REQUIREMENT construct, TEMPLATE construct, or STP.
 
-A deferred type is a deferred argument that can appear as a
-<type-spec> within a REQUIREMENT construct, TEMPLATE construct, or
-STP.
-
 A deferred procedure is a deferred argument that can appear in
 procedure references within a REQUIREMENT construct, TEMPLATE
 construct, or STP.  A deferred procedure's interface shall be
 established in that construct, possibly in terms of other deferred
 arguments.
+
+A deferred type is a deferred argument that can appear as a
+<type-spec>, or in a <declaration-type-spec> or <type-guard-stmt>,
+within a REQUIREMENT construct, TEMPLATE construct, or STP.
+
+Note: A deferred type is not a <type-name> and cannot not appear
+      in a <derived-type-spec>. Consequently a deferred type may
+      not be used in a <structure-constructor>.
 
 Within a construct with a <deferred-arg>, an explicit specification of
 that <deferred-arg> is either <deferred-arg-explicit-stmt> or
@@ -299,40 +303,52 @@ appears as the <function-name> or <subroutine-name> in an
 
 
 <deferred-type-declaration-stmt>
-       <<is>>  TYPE, DEFERRED :: <deferred-type-list>
-       <<or>>  CLASS, DEFERRED :: <deferred-class-list>
+       <<is>>  TYPE, DEFERRED [, ABSTRACT] :: <deferred-type-list>
+       <<or>>  TYPE, ABSTRACT, DEFERRED :: <deferred-type-list>
 
 <deferred-type> <<is>> <name>
 
-<deferred-class> <<is>> <name>
+Constraint: A <deferred-type> entity shall not appear as
+            <parent-type-name> in an EXTENDS attribute.
 
-Constraint: A <deferred-type> or <deferred-class> entity shall not
-            appear as <parent-type-name> in an EXTENDS attribute.
+Constraint: Each <deferred-type> shall appear in <deferred-arg-list>
+            of the innermost scoping unit.
 
-Constraint: Each <deferred-type> shall appear in
-            <deferred-arg-list> of the innermost scoping unit.
+A <deferred-type> is a type that is a deferred argument.
 
-Constraint: Each <deferred-class> shall appear in
-            <deferred-arg-list> of the innermost scoping unit.
+If ABSTRACT appears in <deferred-type-declaration-stmt>, the
+declaration specifies the ABSTRACT attribute for the deferred types.
 
-A <deferred-type> is a deferred type. A <deferred-class> is an
-abstract derived type.
+Note: If a deferred type has the ABSTRACT attribute, its corresponding
+      instantiation argument must be an extensible derived type. 
+      This note is formally specified by a constraint in the 2nd paper.
 
-Note: <deferred-type> will not be allowed in a CLASS declaration, and
-      a <deferred-class> will not be allowed in a TYPE declaration.
-      These are implied by extensions planned for section 7.3.2.1 Type
-      specifier syntax.
+If a deferred type does not have the ABSTRACT attribute, it acts as
+if its instantiation argument were non-extensible.
 
-      Simple example illustrating the above.
+Note: The distinction between abstract and non-abstract deferred types
+      is to ensure a processor can verify the usages of deferred types
+      within a template or requirement are compatible with any
+      instantiation arguments they may have.
+
+      The instantiation argument for an abstract deferred type does
+      not need to itself be abstract, but usage of that deferred type
+      within the template or requirement will act as if the
+      instantiation type were abstract.
+
+      The result is that an abstract deferred type may appear in a
+      CLASS declaration but not a TYPE declaration, and a non-abstract
+      deferred type may appear in a TYPE declaration but not a CLASS
+      declaration. An example:
 
       TEMPLATE TMPL(T, U)
          TYPE, DEFERRED :: T
-         CLASS, DEFERRED :: U
+         TYPE, ABSTRACT, DEFERRED :: U
       CONTAINS
          SUBROUTINE S(X, Y, Z, W)
-            TYPE(T) :: X ! ok
+            TYPE(T) :: X  ! ok
             CLASS(T) :: Y ! invalid
-            TYPE(U) :: Z ! invalid
+            TYPE(U) :: Z  ! invalid
             CLASS(U) :: W ! ok
          END SUBROUTINE S
       END TEMPLATE TMPL
@@ -438,15 +454,14 @@ Constraint: If any ultimate specification of a deferred argument is a
             deferred type, then all ultimate specifications of that
             deferred argument shall be deferred type.
 
-Constraint: If any ultimate specification of a deferred argument is a
-            deferred class, then all ultimate specifications of that
-            deferred argument shall be deferred class.
+Constraint: If any ultimate specification of a deferred type is
+            abstract, then all ultimate specifications of that
+            deferred type shall be abstract.
 
-If any ultimate specification is deferred type, then that argument is
-deferred type.
-
-If any ultimate specification is deferred class, then that argument is
-deferred class.
+If any ultimate specification of a deferred argument is deferred type,
+then that argument is deferred type. If any ultimate specification of
+a deferred type has the ABSTRACT attribute, then that argument has the
+ABSTRACT attribute.
 
 3.3 Deferred argument association
 ---------------------------------
--- a/J3-Papers/syntax-intro-and-deferred-args.txt
+++ b/J3-Papers/syntax-intro-and-deferred-args.txt
@@ -316,30 +316,15 @@ Constraint: Each <deferred-type> shall appear in <deferred-arg-list>
 
 A <deferred-type> is a type that is a deferred argument.
 
-If ABSTRACT appears in <deferred-type-declaration-stmt>, the
-declaration specifies the ABSTRACT attribute for the deferred types.
-
-Note: If a deferred type has the ABSTRACT attribute, its corresponding
-      instantiation argument must be an extensible derived type. 
-      This note is formally specified by a constraint in the 2nd paper.
-
-If a deferred type does not have the ABSTRACT attribute, it acts as
-if its instantiation argument were non-extensible.
+If ABSTRACT appears in <deferred-type-declaration-stmt>, the declared
+deferred types act as if they were abstract extensible derived types.
+Otherwise, they act as if they were non-extensible types.
 
 Note: The distinction between abstract and non-abstract deferred types
-      is to ensure a processor can verify the usages of deferred types
-      within a template or requirement are compatible with any
-      instantiation arguments they may have.
-
-      The instantiation argument for an abstract deferred type does
-      not need to itself be abstract, but usage of that deferred type
-      within the template or requirement will act as if the
-      instantiation type were abstract.
-
-      The result is that an abstract deferred type may appear in a
-      CLASS declaration but not a TYPE declaration, and a non-abstract
-      deferred type may appear in a TYPE declaration but not a CLASS
-      declaration. An example:
+      helps to ensure a processor can verify a template is internally
+      consistent.  For example, a deferred type must not be permitted
+      in a CLASS declaration if it might be instantiated as INTEGER.
+      The following is a simple usage example:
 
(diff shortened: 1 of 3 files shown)


everythingfunctional:
@aury6623 , I expect this to subsume #120 , but let me know what you think.

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -418,15 +450,94 @@ Constraint: If any ultimate specification of a deferred argument is a
...
+      REQUIRES R1{T} ! valid, non-extensible
+   END TEMPLATE
+
+   TEMPLATE T2{T}
+      REQUIRES R2{T} ! valid, extensible
+   END TEMPLATE
+
+   TEMPLATE T3{T}
+      REQUIRES R3{T} ! valid, but T is just extensible
+   END TEMPLATE
```
everythingfunctional, line 480: This is specifically where I'm questioning.

J3-Papers/syntax-template-and-instantiate.txt
```diff
@@ -389,30 +389,36 @@ Note: The previous two constraints constitute what is referred to as
...
-            <deferred-type> shall not be abstract.
+Constraint: An <instantiation-arg> that corresponds to a deferred type
+            that does not have the ABSTRACT attribute shall not be
+            abstract.
 
-Constraint: If an <instantiation-arg> is associated with a
-            <deferred-type>, a variable of that type shall be
-            permitted in a variable definition context.
+Constraint: An <instantiation-arg> that corresponds to a deferred type
+            that does not have the abstract attribute shall be permitted
```
aury6623, line 399:
```suggestion
            that does not have the ABSTRACT attribute shall be permitted
```

aury6623 (commented):
I think this is pretty good

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -298,44 +302,72 @@ appears as the <function-name> or <subroutine-name> in an
...
 
-<deferred-type-declaration-stmt>
-       <<is>>  TYPE, DEFERRED :: <deferred-type-list>
-       <<or>>  CLASS, DEFERRED :: <deferred-class-list>
+<deferred-type-declaration-stmt> <<is>> 
+         TYPE, [<deferred-type-attr-list>] DEFERRED 
+         [, <deferred-type-attr-list>] :: <deferred-type-list>
+
+<deferred-type-attr> <<is>> ABSTRACT
+                     <<or>> EXTENSIBLE
```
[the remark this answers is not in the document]
everythingfunctional: I think I like your suggestion here. One minor nit though is that then the <deferred-type-attr-list> is not optional (i.e. no `[ ]`). I'll fix it.

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -298,44 +302,80 @@ appears as the <function-name> or <subroutine-name> in an
       - class(T) :: local
 
 
-<deferred-type-declaration-stmt>
-       <<is>>  TYPE, DEFERRED :: <deferred-type-list>
-       <<or>>  CLASS, DEFERRED :: <deferred-class-list>
+<deferred-type-declaration-stmt> <<is>> 
+         TYPE, [<deferred-type-attr-list>] :: <deferred-type-list>
```
everythingfunctional, line 306:
```suggestion
         TYPE, <deferred-type-attr-list> :: <deferred-type-list>
```

aury6623 (approved):

J3-Papers/syntax-intro-and-deferred-args.txt
everythingfunctional: Based on our discussion, included that if all specs of a deferred type are abstract that it is abstract.

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -458,11 +458,13 @@ Constraint: If any ultimate specification of a deferred argument is a
...
             deferred argument shall be deferred type.
 
-UTI: Is this how we want this to work?
-
 If any ultimate specification of a deferred type has the EXTENSIBLE
 attribute, then the deferred type has the EXTENSIBLE attribute.
 
+If all ultimate specifications of a deferred type have the ABSTRACT
+attribute, then the deferred type has the ABSTRACT attribute. Otherwise
+the deferred type does not have the ABSTRACT attribute.
```
aury6623, line 466: Should this be a constraint? Otherwise there's nothing to indicate that
```f90
  TEMPLATE T10{T}
      TYPE, DEFERRED, ABSTRACT :: T
      REQUIRES R1{T} ! invalid, T is abstract
   END TEMPLATE
```
is actually *invalid*. It just says that T (in the scope of T10) is not abstract; that's not necessarily an error. But we want the compiler to error here.

```suggestion
Constraint: If any ultimate specification of a deferred type has the
            ABSTRACT attribute, then all ultimate specifications of that
            deferred type shall have the ABSTRACT attribute.
```

everythingfunctional: We do have the necessary constraint in the other paper:

```
Constraint: An <instantiation-arg> that corresponds to a deferred type
            that does not have the ABSTRACT attribute shall not be
            abstract.
```

When taken together with the statement "A deferred type with the ABSTRACT is an abstract derived type."

We don't want to do it the way you wrote in your suggestion because it would make other examples invalid, namely

```
   TEMPLATE T9{T}
      TYPE, DEFERRED, EXTENSIBLE :: T
      REQUIRES R3{T} ! valid, T is just extensible
   END TEMPLATE
```
and
```
   TEMPLATE T15{T}
      REQUIRES R2{T}
      REQUIRES R3{T} ! valid, T is just extensible
   END TEMPLATE
```

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -298,44 +302,80 @@ appears as the <function-name> or <subroutine-name> in an
...
+A <deferred-type> is a deferred type.
 
-Note: <deferred-type> will not be allowed in a CLASS declaration, and
-      a <deferred-class> will not be allowed in a TYPE declaration.
-      These are implied by extensions planned for section 7.3.2.1 Type
-      specifier syntax.
+A deferred type with the EXTENSIBLE attribute is an extensible derived
+type.A deferred type with the ABSTRACT is an abstract derived type. A
+deferred type with the ABSTRACT attribute implicitly has the
+EXTENSIBLE attribute.
```
everythingfunctional, line 333:
```suggestion
A deferred type with the EXTENSIBLE attribute is an extensible derived
type.A deferred type with the ABSTRACT attribute is an abstract derived
type. A deferred type with the ABSTRACT attribute implicitly has the
EXTENSIBLE attribute.
```

J3-Papers/syntax-intro-and-deferred-args.txt
aury6623: Ah, I agree my constraint breaks examples T9 and T15. However, I don't think that constraint you referenced solves the issue. Actually, I think the/another problem here is that the normative text you wrote here creates conflicting assignment of the ABSTRACT attribute.

Above on line 331, there's this statement (with a grammatical error that should be fixed :) I put the fixed word in brackets)

> A deferred type with the ABSTRACT _[attribute]_ is an abstract derived type. 

(Also, we don't actually specify that the EXTENSIBLE and ABSTRACT keywords in the declaration give the extensible or abstract attribute. Also the standard doesn't call "extensible" an "attribute"... ok that part above on line 331 needs another look. But we know what we mean! I think...)

I'm assuming that we intended the paragraph at line 331 to say that the ABSTRACT keyword in a deferred type declaration gives the deferred type the abstract attribute.

But then you wrote

> If all ultimate specifications of a deferred type have the ABSTRACT
attribute, then the deferred type has the ABSTRACT attribute. **Otherwise
the deferred type does not have the ABSTRACT attribute.**

So in T10, the part from line 331 says T is abstract, but your new text says it's not abstract.

aury6623: Oops, I didn't see you already fixed the grammar issue in the next commit :)

everythingfunctional: > So in T10, the part from line 331 says T is abstract, but your new text says it's not abstract.

Actually T10 is invalid, as mentioned in the comment, because of the above constraint.

aury6623: I know T10 is invalid. But again, your new normative text here says that T does not have the abstract attribute. That conflicts with its declaration.

everythingfunctional: :thinking: I think I see what you're getting act. You're saying that my normative text actually makes T11 valid, because since not all ultimate specifications of T have the ABSTRACT attribute, T does not have the ABSTRACT attribute, and therefore could be an instantiation arg for R2. I think I can rework to fix that.

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -418,15 +458,96 @@ Constraint: If any ultimate specification of a deferred argument is a
...
-            deferred class, then all ultimate specifications of that
-            deferred argument shall be deferred class.
+If any ultimate specification of a deferred type has the EXTENSIBLE
+attribute, then the deferred type has the EXTENSIBLE attribute.
 
-If any ultimate specification is deferred type, then that argument is
-deferred type.
+If all ultimate specifications of a deferred type have the ABSTRACT
+attribute, then the deferred type has the ABSTRACT attribute. Otherwise
+the deferred type does not have the ABSTRACT attribute.
```
everythingfunctional, line 466:
```suggestion
If all ultimate specifications of a deferred type have the ABSTRACT
attribute, then the deferred type has the ABSTRACT attribute. Otherwise
if it does not have an explicit specification or has an explicit
specification that does not have the ABSTRACT attribute, then it
does not have the ABSTRACT attribute.
```

aury6623: Why can't you just remove the Otherwise clause entirely? So this paragraph can give the deferred type the abstract attribute, but not take it away.

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -298,44 +302,80 @@ appears as the <function-name> or <subroutine-name> in an
...
+A <deferred-type> is a deferred type.
 
-Note: <deferred-type> will not be allowed in a CLASS declaration, and
-      a <deferred-class> will not be allowed in a TYPE declaration.
-      These are implied by extensions planned for section 7.3.2.1 Type
-      specifier syntax.
+A deferred type with the EXTENSIBLE attribute is an extensible derived
+type.A deferred type with the ABSTRACT attribute is an abstract derived
+type. A deferred type with the ABSTRACT attribute implicitly has the
+EXTENSIBLE attribute.
```
aury6623, line 333: I'm wondering if there needs to be another paragraph here like

> If the ABSTRACT keyword appears, the deferred type has the ABSTRACT attribute. If the EXTENSIBLE keyword appears, the deferred type has the EXTENSIBLE attribute.

The standard seems inconsistent about how it does this for other things. I'm trying to find in the standard where a keyword such as "ABSTRACT" or "BIND(C)" actually gives the entity the ABSTRACT or BIND(C) attributes.

The standard says this in 8.5.1:

> An attribute may be explicitly specified by an _attr-spec_ in a type declaration statement or by an attribute specification statement (8.6). 

That only applies to data objects, though; _attr-spec_ and attribute specification statements are all only for data objects.

For derived types, there is this sentence which makes the SEQUENCE statement give the derived type the SEQUENCE attribute:

> If the SEQUENCE statement appears, the type has the SEQUENCE attribute and is a sequence type.

But I can't find anything like that for ABSTRACT, or any of the things in `binding-attr` (e.g. DEFERRED, NON_OVERRIDABLE). It seems like the standard just implies that keywords in an "attr" BNF as part a declaration give the declared entities those attributes. But then why would it need to specify that specifically for _attr-spec_ in 8.5.1?

J3-Papers/syntax-intro-and-deferred-args.txt
aury6623: Also, either way, do order of declarations matter here? With the constraint you cited,

> Constraint: An <instantiation-arg> that corresponds to a deferred type
            that does not have the ABSTRACT attribute shall not be
            abstract.

it makes sense if the explicit specification comes before the REQUIRE. The T within R1 doesn't have the abstract attribute, so when the explicit specification of T in T10 is passed to it, the compiler errors.

But what if it comes after?
``
  TEMPLATE T10{T}
      REQUIRES R1{T} ! invalid, T is abstract
      TYPE, DEFERRED, ABSTRACT :: T
   END TEMPLATE
```

Does the constraint still apply? Does the compiler need to scan ahead for the explicit specification first, and then recurse into the REQUIRES?

File: J3-Papers/syntax-intro-and-deferred-args.txt (hunk 8 above)
Comment (everythingfunctional) [reply J3-Papers/syntax-intro-and-deferred-args.txt:333]: I think I've got the text correct now, and added a note to clarify the consequence of the rules.

Comment (everythingfunctional) [reply J3-Papers/syntax-intro-and-deferred-args.txt:333]: > do order of declarations matter here?

> Does the compiler need to scan ahead for the explicit specification first,

I don't think the look ahead method would be required, but it does need to do a comparison of any ultimate specifications in effect when it gets to an explicit specification. This will apply for several aspects of deferred arguments.

Comment (aury6623) [reply J3-Papers/syntax-intro-and-deferred-args.txt:333]: Both of your responses on this comment thread are for a different comment thread 😄 But okay, I think the order of declarations thing is workable. And I think your new text is okay. (I'm still not sure why just "If all ultimate specifications of a deferred type have the ABSTRACT attribute, then the deferred type has the ABSTRACT attribute." doesn't work. But what you have seems to work too.)

File: J3-Papers/syntax-intro-and-deferred-args.txt [hunk 9]
```diff
@@ -298,44 +302,80 @@ appears as the <function-name> or <subroutine-name> in an
...
 
-Constraint: A <deferred-type> or <deferred-class> entity shall not
-            appear as <parent-type-name> in an EXTENDS attribute.
+Constraint: The same <deferred-type-attr> shall not appear more than
+            once in a given <deferred-type-attr-list>.
 
-Constraint: Each <deferred-type> shall appear in
-            <deferred-arg-list> of the innermost scoping unit.
+Constraint: ABSTRACT and EXTENSIBLE shall not both appear in the same
+            <deferred-type-attr-list>.
```
everythingfunctional, line 318:
```suggestion
```

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -298,44 +302,80 @@ appears as the <function-name> or <subroutine-name> in an
...
+A <deferred-type> is a deferred type.
 
-Note: <deferred-type> will not be allowed in a CLASS declaration, and
-      a <deferred-class> will not be allowed in a TYPE declaration.
-      These are implied by extensions planned for section 7.3.2.1 Type
-      specifier syntax.
+A deferred type with the EXTENSIBLE attribute is an extensible derived
+type.A deferred type with the ABSTRACT attribute is an abstract derived
+type. A deferred type with the ABSTRACT attribute implicitly has the
+EXTENSIBLE attribute.
```
everythingfunctional, line 333:
```suggestion
type. A deferred type with the ABSTRACT attribute implicitly has the
EXTENSIBLE attribute, which can be confirmed with an explicit inclusion
of the EXTENSIBLE keyword.
```

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -300,41 +304,81 @@ appears as the <function-name> or <subroutine-name> in an
...
+
+Constraint: The same <deferred-type-attr> shall not appear more than
+            once in a given <deferred-type-attr-list>.
 
 Constraint: Each <deferred-type> shall appear in <deferred-arg-list>
             of the TEMPLATE, REQUIREMENT or STP in which it appears.
 
-<deferred-type> <<is>> <name>
+Constraint: Each <deferred-type> shall appear in <deferred-arg-list>
+            of the TEMPLATE, REQUIREMENT or STP in which it appears.
```
everythingfunctional, line 323:
```suggestion
```

J3-Papers/syntax-intro-and-deferred-args.txt
```diff
@@ -300,41 +304,79 @@ appears as the <function-name> or <subroutine-name> in an
...
+
+Constraint: DEFERRED shall appear in each <deferred-type-attr-list>.
+
+Constraint: The same <deferred-type-attr> shall not appear more than
+            once in a given <deferred-type-attr-list>.
 
 Constraint: Each <deferred-type> shall appear in <deferred-arg-list>
             of the TEMPLATE, REQUIREMENT or STP in which it appears.
 
-<deferred-type> <<is>> <name>
```
everythingfunctional, line 322:
```suggestion
```

aury6623 (approved):

PR #132, opened 2024-05-23
minor edits/corrections

everythingfunctional:


J3-Papers/syntax-template-and-instantiate.txt
```diff
@@ -399,10 +399,6 @@ Constraint: If an <instantiation-arg> is associated with a
             <deferred-type>, a variable of that type shall be
             permitted in a variable definition context.
 
-Constraint: If an <instantiation-arg> is a <type-spec>, it shall not
-            specify a type that has a coarray potential subobject
-            component.
-
```
everythingfunctional, line 405: I'm convinced this is unnecessary. Intrinsic assignment doesn't cause coarray (re)allocation.

everythingfunctional: On second thought, I think we should still leave it in.

J3-Papers/syntax-template-and-instantiate.txt
```diff
@@ -399,10 +399,6 @@ Constraint: If an <instantiation-arg> is associated with a
             <deferred-type>, a variable of that type shall be
             permitted in a variable definition context.
```
everythingfunctional, line 401:
```suggestion

Constraint: If an <instantiation-arg> is a <type-spec>, it shall not
            specify a type that has a coarray potential subobject
            component.
```

J3-Papers/syntax-template-and-instantiate.txt
```diff
@@ -402,7 +402,6 @@ Constraint: If an <instantiation-arg> is associated with a
 Constraint: If an <instantiation-arg> is a <type-spec>, it shall not
             specify a type that has a coarray potential subobject
             component.
-
 Constraint: If an <instantiation-arg> corresponds to <deferred-class>,
```
everythingfunctional, line 405:
```suggestion

Constraint: If an <instantiation-arg> corresponds to <deferred-class>,
```

PR #138, opened 2024-07-29
Write a paper to change instantiation of STPs

everythingfunctional:


tclune (approved):
Reads good to me.   Very curious as to Malcolm's opinion on the options.

aradi:
I know, it is merged already, but wouldn't
```fortran
INSTANTIATE new_name => templ_proc(...)
```
be the most comsistent with how we use renaming in Fortran already (when importing modules)?

PR #140, opened 2024-08-12
write a paper solving long template arg list problem

everythingfunctional:
This proposes a feature and some possible syntax to help alleviate the problem of long template argument lists.


everythingfunctional:
@certik , this is the paper I was showing earlier today. I tried to simplify the example just a bit to hopefully make it easier to grok.

certik:
Awesome, thanks for this!

PR #144, opened 2024-10-21
Formatting, mostly indent

aury6623:
Well 161 and 162 are already up for vote, so maybe we shouldn't change those yet. If/when they need a revision, we can possibly apply these formatting changes to them too. But for the others, we can apply them now, if you all agree.

This fixes some tabs and alignment of constraint text. The bulk of the formatting is in the misc edits paper, where I try to make the wording and formatting more consistent across all the many small edits in that paper.


aury6623:
I'm not allowed to request a review apparently, so I'll just @ you two: @everythingfunctional  @tclune

everythingfunctional (approved):
Nearly all whitespace changes, and effectively no content changes, so seems fine to me.

tclune (approved):

PR #145, opened 2024-10-22
Replace proposed C501 with constraints on <template> and <requirement>

aury6623:
I based the new constraints off of C873, which limits the accessibility statement to only modules.

Also, I noticed C2057 (now C2057b to make room for the new constraint) was a bit vague, so I updated it to be more explicit about _how_ the \<deferred-arg\> needs to appear.

@tclune @everythingfunctional

### DIFF
--- a/J3-Papers/edits/24-163_templates.txt
+++ b/J3-Papers/edits/24-163_templates.txt
@@ -26,14 +26,17 @@ R2010 <template> <<is>> <template-stmt>
                       [ <template-subprogram-part> ]
                       <end-template-stmt>
 
+C2027a (R2010). A <template> shall only appear in the <specification-part>
+                of a main program or module.
+
 R2011 <template-stmt> <<is>>
           TEMPLATE <template-name> ([<deferred-arg-list>])
 
 R2012 <end-template-stmt> <<is>>  END [TEMPLATE [<template-name>]]
 
-C2027 (R2012). If a <template-name> appears in the <end-template-stmt>,
-            it shall be identical to the <template-name>
-            specified in the <template-stmt>.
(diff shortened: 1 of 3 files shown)


everythingfunctional (approved):
Seems straightforward enough to me, but I want to let Tom take a look too.

tclune (commented):
I have no problem with the change as such,  but did we really mean to disallow 
 defining a REQUIREMENT within a TEMPLATE construct?
 
 I would have thought we have examples that violate this.

everythingfunctional:
> did we really mean to disallow
> defining a REQUIREMENT within a TEMPLATE construct?

Maybe not intentionally, but I don't know of a good example of why you would want to do that.

aury6623:
One thing before merging -- the changed files' names have the J3 paper number in them, but by merging this JR they will be different from what's uploaded. I don't want to change them to "r1" or something preemptively, since they haven't been reuploaded yet. Should I change the names to something like `24-163_draftr1_templates.txt`?

everythingfunctional:
Maybe @tclune should upload it, then you can change the name and then merge? This is clearly not the smoothest workflow.

tclune:
@everythingfunctional Not sure what you are asking for.

When the dust settles (PRs merged) what do we want the directory to look like?   2 papers one called r1?    Once I understand the end game I can make it happen.

Am guessing the easiest path for now is to approve this PR and then fix up after.

everythingfunctional:
End state is the repo contains just 1 file, which contains and is named for the version voted on. That means if we agree that this is the version that should be submitted you can upload it now, because whether the name gets changed before or after merging the PR is somewhat immaterial at that point.

PR #148, opened 2024-11-12
Changes from Nov 11 meeting

aury6623:


J3-Papers/edits/24-164_instantiate.txt
```diff
@@ -202,10 +205,9 @@ second deferred argument in the reduced list, etc.
 R1529 <instantiation-arg> <<is>> <constant-expr>
                           <<or>> <type-spec>
                           <<or>> <generic-spec>
-                          <<or>> <procedure-name>
```
aury6623, line 205: I changed `<procedure-name>` to `<specific-procedure>` because it already has a good constraint in the standard:
```
R1507 <specific-procedure> is <procedure-name>
...
C1508 (R1507) A <procedure-name> shall denote a nonintrinsic procedure that has an explicit interface.
```
I believe that is accurate for what we want here.

aury6623:
@everythingfunctional @tclune All the stuff we talked about today, except one thing that we didn't talk about, which I mentioned in the previous comment.

everythingfunctional (approved):

J3-Papers/edits/24-163_templates.txt
```diff
@@ -151,54 +157,20 @@ R1521 <templated-subprogram-declaration-construct>
...
-      interfaces, operators, or intrinsic assignment.
+C1535 A procedure referenced anywhere within a template or templated
+      subprogram shall have an explicit interface.
 
 NOTE
 
-    This constraint expresses the notion of type-safety.  In particular,
+    Constraint C1535 expresses the notion of type-safety.  In particular,
     this allows a template implementation to be verified to be internally
     consistent.
```
tclune, line 175: Low priority - but I think this could be worded better by turning it around.  Something like,  ... allows the implementor to verify that a template in internally consistent"

tclune (approved):

PR #175, opened 2026-04-13
Feature/tclune uti009 etc

tclune:


J3-Papers/utis/syntax-correction-deferred-args.txt
everythingfunctional, line 1: I saw a few other notes/examples that have '()' instead of '{}'. Are we sure all of those handled by other UTIs?

J3-Papers/utis/UTI010.txt
```diff
@@ -0,0 +1,108 @@
...
+  DO i = 1, n-1
+    iterate = f(iterate)
+  END DO
+END FUNCTION iterate
+
+INTEGER :: m
+REAL :: y
+
+m = iterate{INTEGER,7,increment}(10) ! yields 17
+y = iterate{REAL,4,times2}(7.)       ! yields 102
```
everythingfunctional, line 106: Inline instantiation is covered after this. Do we want to include this example here, or add another note and example later for this?

J3-Papers/utis/syntax-correction-deferred-args.txt
tclune: Will find those as we go and/or in the ultimate fix {} paper.

PR #177, opened 2026-04-13
draft response to uti 12

everythingfunctional:


J3-Papers/utis/uti012-response.txt
```diff
@@ -0,0 +1,95 @@
...
+    Perhaps you want a CODIMENSIONABLE or COINDEXABLE
+    deferred-type-attr?
+
+    I further note that as written, there can be at most one 
+    deferred-type-attr and thus no need for a list or the constraints...
+    that would, of course, change if we did add
+    CODIMENSIONABLE/COINDEXABLE.
+
+To the extent that templates are not interoperating with coarrays
+they are also not interoperating with MPI. For example the following
```
tclune: I do not understand why this example is invalid?   Is it that we can't have ALLOCATABLE attribute on BUFF?  Or something subtle in the MPI_ISEND() interface that I'm not immediately seeing?

everythingfunctional: Now that I am looking more closely at this, I think we need to do something about disallowing deferred type arguments to `TYPE(*)` dummies. I recently found out that you can't pass derived-types with length type parameters to a `TYPE(*)` dummy argument in a `BIND(C)` procedure. I don't know that between the MPI standard and the Fortran standard if it explicitly forbids it, since the MPI standard doesn't necessarily say that the Fortran interfaces are `BIND(C)`. Of course there's almost certainly `BIND(C)` in there somewhere, and that's where it becomes non-standard.

everythingfunctional: Found the actual sentence:

> 15.5.2.5 Ordinary dummy variables
> ...
> If the actual argument is of a derived type that has type parameters, type-bound procedures, or final subroutines, the dummy argument shall not be assumed-type.

tclune: Does this mean one can't pass character data to `TYPE(*)`?   Surprises me, but only a bit I suppose.

tclune (approved):

PR #203, opened 2026-06-01
Fixes from HPC subgroup

tclune:
HPC subgroup approached me about UTI007 and how it relates to auto generics (generic subgrograms).

As it turns out, I failed to upload this paper in my big sweep, so it is good that they noticed it.

### DIFF
--- a/J3-Papers/utis/uti007-response.txt
+++ b/J3-Papers/utis/uti007-response.txt
@@ -41,7 +41,7 @@ point. Edits are provided.
 
 
 
-[24:20-22] delete definition of "same statement". Also delete UTI 007.
+[24:22] Delete UTI 007.
 
 [172:16] hyperlink "same statement" to new subclause 16.5.4.1 and add
          reference to the subclause.
--- a/J3-Papers/utis/uti007-response.txt
+++ b/J3-Papers/utis/uti007-response.txt
@@ -1,5 +1,5 @@
 To: J3                                                     J3/##-###
-From: Brad Richardson & Generics
+From: HPC & Generics
 Subject: Response to UTI007
 Date: 
 
@@ -35,6 +35,21 @@ In reference to new definition:
 Yes, this was an attempt to handle image control statements appearing
 in templates. The suggestion that we do this a better way is a valid
 point. Edits are provided.
+
+[24:21+] 3.121 same statement
+
+    Replace ", is from the same instance of the template" with "."
+
+    Append 2 notes:
+
+    Note 1: If the statement is within a template it is from the same
+            instance.
+    <cross-reference to new 16.5.4.1 Same statement requirements>
+  
+    Note 2: If the statement is within a generic subprogram, it is
+            from the same specific procedure.
+    <cross-reference to ??(new) in generic subprogram>
+
 
 [24:20-22] delete definition of "same statement". Also delete UTI 007.
 
--- a/J3-Papers/utis/uti007-response.txt
+++ b/J3-Papers/utis/uti007-response.txt
@@ -39,19 +39,6 @@ appearing in generic subprograms.
 The suggestion that we do this a better way is a valid
 point. Edits are provided.
 
-[24:21+] 3.121 same statement
-
-    Replace ", is from the same instance of the template" with "."
-
-    Append 2 notes:
-
-    Note 1: If the statement is within a template it is from the same
-            instance.
-    <cross-reference to new 16.5.4.1 Same statement requirements>
-  
-    Note 2: If the statement is within a generic subprogram, it is
-            from the same specific procedure.
-    <cross-reference to ??(new) in generic subprogram>
 
 
 [24:20-22] delete definition of "same statement". Also delete UTI 007.
--- a/J3-Papers/utis/uti007-response.txt
+++ b/J3-Papers/utis/uti007-response.txt
@@ -32,7 +32,7 @@ In reference to new definition:
     the statement is within a template, is from the same instance of
     the template
 
-Yes, this was an attempt to handle image control statements appearing
+Yes, this was an attempt to handle image control statements and collective subroutines appearing
 in templates. The suggestion that we do this a better way is a valid
 point. Edits are provided.
 
--- a/J3-Papers/utis/uti007-response.txt
+++ b/J3-Papers/utis/uti007-response.txt
@@ -33,7 +33,10 @@ In reference to new definition:
     the template
 
 Yes, this was an attempt to handle image control statements and collective subroutines appearing
-in templates. The suggestion that we do this a better way is a valid
+in templates. In addition, it was realized that a similar problem also 
+applies to invocation of statements with "same statement" requirements
+appearing in generic subprograms.
+The suggestion that we do this a better way is a valid
 point. Edits are provided.
 
 [24:21+] 3.121 same statement


J3-Papers/utis/uti007-response.txt
```diff
@@ -36,6 +36,21 @@ Yes, this was an attempt to handle image control statements appearing
...
+
+    Append 2 notes:
+
+    Note 1: If the statement is within a template it is from the same
+            instance.
+    <cross-reference to new 16.5.4.1 Same statement requirements>
+  
+    Note 2: If the statement is within a generic subprogram, it is
+            from the same specific procedure.
+    <cross-reference to ??(new) in generic subprogram>
```
bonachea: In following this plan, we obviously also need an edit to introduce a new subclause for generic subprogram near 15.6.2.4 that we reference here.

J3-Papers/utis/uti007-response.txt
```diff
@@ -36,6 +36,21 @@ Yes, this was an attempt to handle image control statements appearing
...
+
+
 [24:20-22] delete definition of "same statement". Also delete UTI 007.
```
bonachea:
```suggestion
[24:22] Delete UTI 007.
```

J3-Papers/utis/uti007-response.txt
bonachea:
```suggestion
Yes, this was an attempt to handle image control statements and collective subroutines appearing
```

J3-Papers/utis/uti007-response.txt
```diff
@@ -36,6 +36,21 @@ Yes, this was an attempt to handle image control statements appearing
 in templates. The suggestion that we do this a better way is a valid
```
bonachea:
```suggestion
in templates. In addition, it was realized that a similar problem also 
applies to invocation of statements with "same statement" requirements
appearing in generic subprograms.
The suggestion that we do this a better way is a valid
```

bonachea (changes requested):
Thanks for working on this @tclune and @everythingfunctional .

Here are my first round of comments

J3-Papers/utis/uti007-response.txt
```diff
@@ -36,6 +36,21 @@ Yes, this was an attempt to handle image control statements appearing
...
+
+    Note 1: If the statement is within a template it is from the same
+            instance.
+    <cross-reference to new 16.5.4.1 Same statement requirements>
+  
+    Note 2: If the statement is within a generic subprogram, it is
+            from the same specific procedure.
+    <cross-reference to ??(new) in generic subprogram>
+
+
```
everythingfunctional: The following edit is deleting this definition, so no, don't do this. Instead, if you want to also talk about generic subprograms, include it in the new section, 16.5.4.1. Or maybe this is a problem data needs to handle wherever they added things dealing with auto generics.

J3-Papers/utis/uti007-response.txt
bonachea, line 59: All these hyperlinks (and the ones below) should now go to the Glossary entry instead

bonachea, line 58: Also missing [254:14] in the same paragraph

J3-Papers/utis/uti007-response.txt
```diff
@@ -36,6 +36,21 @@ Yes, this was an attempt to handle image control statements appearing
...
 [24:20-22] delete definition of "same statement". Also delete UTI 007.
```
bonachea, line 45: The list of "same statement" hyperlinks below omits some instances of that phrase in the standard. Was this deliberate?

For example [168:36] Image Selectors:

> The denotation of a stat-variable in an image-selector shall not depend on the evaluation of any entity in the **same statement**. The value of an expression shall not depend on the value of any stat-variable that appears in the **same statement**.

IIUC this paragraph is attempting to forbid code such as:
```fortran
x = A[1, stat=S(foo())] + A[1, stat=SS] + SS
```
This particular case is not talking about matching statements across images, but it might still be relevant to link the definition here?

bonachea: Another example from 10.1.4 Evaluation of operations:

[184:6-7]
> if a function reference causes definition or undefinition of an actual argument of the function, that argument or any associated entities shall not appear elsewhere in the **same statement.**
> 
> NOTE 1
For example, the statements
> ```
> A (I) = F (I)
> Y = G (X) + X
> ```
> are prohibited if the reference to F defines or undefines I or the reference to G defines or undefines X.

Here again the standard is not talking about matching statements across images, but by not hyperlinking to the Glossary term are we implying there is some other/different meaning to the phrase "same statement" here?

bonachea: The fact the "same statement" phrase appears in a number of places in the standard that have nothing to do with collectively-invoked multi-image features _might_ argue for minting an entirely new term (e.g. perhaps "matched statement"?) that is specific to matching of statements across images.

Then we'd instead define that term in the glossary and/or subclauses, and change all the current occurrences of "same statement" relating to collectively-invoked multi-image  features to instead use the new term and hyperlink the definition. Under this plan we'd deliberately leave the some instances of "same statement" (such as those quoted above) unchanged and unlinked.

Thoughts?

everythingfunctional: These examples are the reason I was sympathetic with Malcolm's point that we shouldn't redefine a normal english phrase. My inclination, as with adding the subclause below, was just to point out the context in which we need the additional nuance. I think the below subclause is sufficient for templates. It may be useful to add something similar for autogenerics, but the description of autogenerics may already sufficiently make the implications clear enough.

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,10 +32,28 @@ In reference to new definition:
...
+
+    Append 2 notes:
+
+    Note 1: If the statement is within a template it is from the same
+            instance.
+    <cross-reference to new 16.5.4.1 Same statement requirements>
+  
+    Note 2: If the statement is within a generic subprogram, it is
+            from the same specific procedure.
+    <cross-reference to ??(new) in generic subprogram>
```
tclune:
```suggestion
```

everythingfunctional (approved):

bonachea:
@tclune the stated goal of this PR was to amend the UTI007 response "as it relates to auto generics (generic subgrograms)", but what you've just merged doesn't appear to do that at all.

In particular, it now mentions the problem with generic subprograms in the introduction, but then provides no edits to solve it.

Are you planning another PR to address this? In particular, I think it's premature to delete UTI007 until this is explicitly addressed with edits.

J3-Papers/utis/uti007-response.txt
bonachea: @everythingfunctional said:
> t may be useful to add something similar for autogenerics, but the description of autogenerics may already sufficiently make the implications clear enough.

I do NOT think the current subclause on generic subprograms makes this clear at all.

Also, even if we "add something similar for autogenerics", the "same statement" phrases can only hyperlink to ONE of the new subclauses, where we cannot know which is relevant to the reader's use case.

J3-Papers/utis/uti007-response.txt
bonachea: This was never addressed, and should not have been marked resolved

tclune:
@bonachea Apologies - our subgroup meeting was a bit rushed yesterday because I had a dental appt. to run to.   Thanks for catching the inconsistency in the edit.   My understanding is that you and @everythingfunctional have now spoken and he'll do another PR for this one.

PR #206, opened 2026-06-04
use a different tactic for uti 007

everythingfunctional:


J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,44 @@ In reference to new definition:
...
 
-[172:16] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[172:16] change "same statement" to "corresponding statement" and link
+         to definition.
 
-[175:38] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[175:38] change "same statement" to "corresponding statement" and link
+         to definition.
```
bonachea, line 62: I just realized there's an important edit missing at [221:31-32], for the sentence that currently reads:

> All active images of the new team shall execute the same CHANGE TEAM statement.

Here F23 is not using the phrase "same statement", but clearly the same meaning is implied. So in order to be a complete solution we have to somehow morph this text to use the new defined term, eg:




```suggestion

[221:31-32] change "the same CHANGE TEAM statement" a "a corresponding CHANGE TEAM statement" and link "corresponding" to definition.
```

or perhaps:


```suggestion

[221:31-32] change "the same CHANGE TEAM statement" a "a CHANGE TEAM statement in a corresponding statement" and link "corresponding statement" to definition.
```

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,44 @@ In reference to new definition:
...
+collective subroutines appearing in templates. In addition, it was
+realized that a similar problem also  applies to invocation of
+statements with "same statement" requirements appearing in generic
+subprograms. The suggestion that we do this a better way is a valid
 point. Edits are provided.
 
+[10:35+] Add "corresponding statement" with the following definition,
+         "same statement that, if within a template is from the same 
+         instance of the template, or if within a generic subprogram
+         is from the same specific procedure."
```
bonachea: I'm in 100% agreement that we should define a new term (that roughly means "textual alignment after separation via template/generic expansion"), and apply/link it everywhere appropriate.

However I have some hesitation with the specific spelling "corresponding statement". In particular, the standard already uses the adjective "corresponding" in image-adjacent situations where it means something that is _often_ but not _always_ the same as what we mean here.

For example from SYNC IMAGES [250:10-15]:

> Execution of a SYNC IMAGES statement performs a synchronization of the image with each of the other images in the image-set. Executions of SYNC IMAGES statements on images M and T **correspond** if the number of times image M has executed a SYNC IMAGES statement in the current team with T in its image set is the same as the number of times image T has executed a SYNC IMAGES statement with M in its image set in this team. The segments that executed before the SYNC IMAGES statement on either image precede the segments that execute after the **corresponding** SYNC IMAGES **statement** on the other image.

This paragraph literally defines what it means for two SYNC IMAGES statements to "correspond" and even uses the phrase "corresponding (SYNC IMAGES) statement". However in this case the term does NOT mean the same thing as the new definition we are proposing. In particular, there is no current "same textual statement" requirement for SYNC IMAGES, nor should we be proposing or implying that new requirement (because it would deeply injure the usability of this feature).

Similarly from SYNC TEAM [252:5-9]:

> Successful execution of a SYNC TEAM statement performs a synchronization of the team identified by team value. Execution on an image, M, of the segment following the SYNC TEAM statement is delayed until each other image of the specified team has executed a SYNC TEAM statement specifying the same team as many times as has image M in this team. The segments that executed before the SYNC TEAM statement on an image precede the segments that execute after the **corresponding** SYNC TEAM **statement** on another image.

Here again there is no "same textual statement" requirement for SYNC TEAM, so it's problematic to imply the adjective "corresponding" as applied to a statement means any form of textual alignment. If we really wanted to spell our new term "corresponding statement", I think we'd also have to go find all such current uses and rephrase them to avoid the overloaded "corresponding" adjective.

As an additional cognitive complication, the standard already widely uses the term "corresponding coarray(s)" to describe a cross-image relationship, but this (in general) is a dynamic property of coarray objects, not a static property of the program text.

**I'd strongly prefer we choose a different term here that avoids these conflicts. Ideally we'd choose a "fresh" term that is not already wildly overloaded for other purposes elsewhere in the standard.**

J3-Papers/utis/uti007-response.txt
bonachea: I'm now also worried we may have missed other instances of cross-image `same statement` requirements that are instead spelled `same [something] statement`, but I don't know how to robustly search for that.

bonachea (changes requested):
Thanks @everythingfunctional for writing up this PR!
As we discussed verbally, I think this is the right general approach to solving the overall problem.

However I have some concerns with the details, that I think should be addressed before we upload this proposal.

J3-Papers/utis/uti007-response.txt
everythingfunctional: I know you had originally proposed "matching", which might be good enough. I initially thought of corresponding as a good word because it seemed to go with the term for coarrays, but it's use in SYNC TEAM probably makes that not the right word to use here. Perhaps "correlated" would work? Any other ideas?

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,45 @@ In reference to new definition:
...
+         to definition.
 
-[175:38] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[175:38] change "same statement" to "corresponding statement" and link
+         to definition.
 
-[221:33] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[221:31-32] change "the same CHANGE TEAM statement" a "a corresponding CHANGE TEAM statement" and link "corresponding" to definition.
```
bonachea: Minor wording/formatting fix.

```suggestion
[221:31-32] change "the same CHANGE TEAM statement" to
            "a corresponding CHANGE TEAM statement" 
            and link "corresponding" to definition.

```

J3-Papers/utis/uti007-response.txt
bonachea: Some other possible alternate spellings for the new phrase that I've thought about:

1. Same instantiated statement
2. Selfsame statement
3. Matching statement


I'm not totally happy with any of these, but at first glance they all seem less problematic than "corresponding statement"

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,48 @@ In reference to new definition:
...
 
-[175:38] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[175:38] change "same statement" to "matching statement" and link
+         to definition.
 
-[221:33] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[221:31-32] change "the same CHANGE TEAM statement" to "a CHANGE
+            TEAM statement in a corresponding statement" and link
```
bonachea: Missed fixing one instance:
```suggestion
            TEAM statement in a matching statement" and link
```

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,48 @@ In reference to new definition:
...
+collective subroutines appearing in templates. In addition, it was
+realized that a similar problem also  applies to invocation of
+statements with "same statement" requirements appearing in generic
+subprograms. The suggestion that we do this a better way is a valid
 point. Edits are provided.
 
+[10:35+] Add "matching statement" with the following definition,
+         "same statement that, if within a template is from the same 
+         instance of the template, or if within a generic subprogram
+         is from the same specific procedure."
```
bonachea: I suggest we additionally add some cross-references here.


```suggestion
         "same statement that, if within a template is from the same 
         instance (3.120) of the template, or if within a generic subprogram
         is from the same specific procedure.
         
         Note 1 to entry: Generic subprograms are described in 15.6.2.4."
         
         Link the phrase "same instance" to the Glossary definition (3.120).
```

NOTE: this comment depends on the resolution to UTI 006 regarding 3.120. 
@everythingfunctional What is the status of UTI 006?

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,48 @@ In reference to new definition:
...
-appearing in generic subprograms.
-The suggestion that we do this a better way is a valid
+Yes, this was an attempt to handle image control statements and
+collective subroutines appearing in templates. In addition, it was
+realized that a similar problem also  applies to invocation of
+statements with "same statement" requirements appearing in generic
+subprograms. The suggestion that we do this a better way is a valid
 point. Edits are provided.
 
+[10:35+] Add "matching statement" with the following definition,
```
bonachea: `[10:35+]` seems like an erroneous placement for the new entry; this is neither alphabetical nor grouped under a meaningful heading.

I think the best choice is probably grouped under "3.134 statement":

```suggestion
[26:37+] Add "3.134.1+ matching statement" with the following definition,
```

bonachea (changes requested):
Thanks for the improvements @everythingfunctional 

This is close to ready, but I noticed a few additional problems and added suggestions to resolve them.

J3-Papers/utis/uti007-response.txt
everythingfunctional: The resolution to UTI 006 is to remove the definition of "same instance", so I'll suggest we just link/reference the subclause in which it is defined.

J3-Papers/utis/uti007-response.txt
everythingfunctional: seems reasonable.

J3-Papers/utis/uti007-response.txt
```diff
@@ -32,53 +32,54 @@ In reference to new definition:
...
+[24:20-22] delete definition of "same statement".
 
 [24:22] Delete UTI 007.
 
-[172:16] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
-
-[175:38] hyperlink "same statement" to new subclause 16.5.4.1 and add
-         reference to the subclause.
+[26:39+] Add "3.134.1+ matching statement" with the following
```
bonachea: "matching statement" should appear alphabetically _before_ "nonexecutable statement"
```suggestion
[26:37+] Add "3.134.1+ matching statement" with the following
```

bonachea (changes requested):
One minor but important typo, otherwise LGTM!

bonachea (approved):

tclune (approved):

